PR #5, opened 2015-10-24
Handle key= and ref=

vvo:


vvo:
cc @mjackson what do you think on this? Seems we need to ensure manual key and ref were added right? Thanks

Issue #9, opened 2015-11-30
Array.prototype.fill


Sinewyk:
Array.prototype.fill polyfill is not ~~standard~~ implemented everywhere yet.

```
'undefined' is not a function (evaluating 'new Array(times).fill('  ')')
```

Make a note about it in the README ?

vvo:
Ah yes I think I made the assumption that everyone was using babel BUT that's not a good assumption. I will import fill from a polyfill. Do you mind trying to do a PR?

vvo:
Something like

```js
import fill from 'lodash/array/fill';
```

Sinewyk:
Oh yeah, nice, forgot lodash is always becoming more awesome.

Sinewyk:
Had a bit of trouble to find it on npm, you have to go to https://www.npmjs.com/browse/keyword/lodash-modularized and search pages by hand it seems :o ...

Couldn't find it from search box :disappointed: 

But here you are.

vvo:
thx! will publish

PR #10, opened 2015-11-30
use fill from lodash

Sinewyk:
because not all VMs have implemented Array.prototype.fill yet
fix https://github.com/algolia/react-element-to-jsx-string/issues/9


vvo:
released as 2.1.3

PR #13, opened 2015-12-10
Fix <div>{''}</div>

smacker:
It is not meaningful child and should be filtered.

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -417,4 +417,10 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
   123
 </div>`);
   });
+
+  it(`reactElementToJSXString(<div>{''}</div>)`, () => {
+    expect(
+      reactElementToJSXString(<div>{''}</div>)
+    ).toEqual(reactElementToJSXString(<div/>));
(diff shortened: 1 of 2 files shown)


vvo:
Not sure, it seems to be a difference between `<div/>` and `<div>{''}</div>`: https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=%3Cdiv%3E%7B''%7D%3C%2Fdiv%3E

smacker:
yes. it generates a little bit different jsx, but same DOM. It's very ugly to write in test:
```
expect(output).toEqualJSX(<div>
{''}
</div>)
```

Proof about DOM:

```
> document.querySelector('body').innerHTML = '<div class="bla"></div>'
< "<div class="bla"></div>"
> React.render(React.createElement('div', null, ''), document.querySelector('.bla'))
< <div data-reactid=​".8">​</div>​
> document.querySelector('.bla').innerHTML
< "<div data-reactid=".8"></div>"
```

vvo:
WFM

PR #14, opened 2016-01-05
Extra new line with numbers

smacker:
Hi! I added a test and a fix for https://github.com/algolia/expect-jsx/issues/14

But I'm not sure that it's correct behavior. Also it's breaking change. Please consider this changes, but don't merge it without bump of major version. (at least my tests will be broken)

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -432,4 +432,12 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
       reactElementToJSXString(<div>{''}</div>)
     ).toEqual(reactElementToJSXString(<div/>));
   });
+
+  it(`reactElementToJSXString(<div>String with {1} js expression</div>)`, () => {
+    expect(
+      reactElementToJSXString(<div>String with {1} js number</div>)
+    ).toEqual(`<div>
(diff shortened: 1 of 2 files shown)


vvo:
Good! I wonder if there are some other types we might want to threat so?

smacker:
Here: https://github.com/facebook/react/blob/3b96650e39ddda5ba49245713ef16dbc52d25e9e/src/shared/utils/traverseAllChildren.js#L106 react check only for strings and numbers. I guess it will be okay if we add new line for any other type.

redox:
:+1: @smacker

vvo:
Awesome, gonna publish this. I still wonder if I should bump major, since this looks more like a undercover bug than a complete major change?

redox:
> I still wonder if I should bump major

I would definitely consider it as a bug.

smacker:
It's up to you. But I guess not only my tests will be broken.

vvo:
We decided to not bump major @smacker since it should not impact TOO many users. As you discovered this was not a good behavior (and the error message should be straightforward).

Thanks A LOT for contributing :)

Issue #15, opened 2016-01-07
Ability to configure how to show the component's display name


leoasis:
Some times you want to have control on how to display components' names. When you have components wrapped in a higher order component, sometimes it makes sense to just hide that when rendering the JSX and just show the original component's name (i.e. when rendering to a string that will be used as the markup example in a styleguide). This should not only be configurable for top level components, but also for all the components appearing as props in any level. I think probably providing a function that receives an element and returns the display name as part of some options to provide to the `reactElementToJSXString` function, would be the right API.

What do you think?

vvo:
Hi @leoasis, indeed that could be a good addition.

I would be happy to review a PR with tests for something like this:

```js
reactElementToJSXString(el, {displayName: fn});
```

When fn would receive the current element that we want to choose a name for.

vvo:
What are you using react-element-to-jsx-string for? Thanks

leoasis:
Cool, I'll look into doing something today or maybe tomorrow!

I'm currently using it for showing the JSX of examples in a live styleguide, aka component catalog. Since I have _themeable_ components, and the theming logic is being abstracted in a higher order component (with the theme styles being passed down as props), what I actually have are `Themed(Component)` kind of display names that I'd want to change when rendering the example JSX string

PR #16, opened 2016-01-11
Add displayName option for custom formatting

leoasis:
Closes https://github.com/algolia/react-element-to-jsx-string/issues/15

This adds the ability to pass a custom function to format the way components are displayed in the final string.

I've had to pass around the `options` object on all the functions used for recursion. This in my opinion is not very clear, I think it may be better to move the small function under `toJSXString` so that the `options` variable is in scope and can be accessed whenever we recurse again from those functions.

If you agree that's clearer and prefer that approach instead of this current one, let me know and I can modify the PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,11 @@ import sortobject from 'sortobject';
 import traverse from 'traverse';
 import fill from 'lodash/array/fill';
 
-export default function reactElementToJSXString(ReactElement) {
-  return toJSXString({ReactElement});
+export default function reactElementToJSXString(ReactElement, options) {
+  return toJSXString({ReactElement, options});
 }
 
-function toJSXString({ReactElement = null, lvl = 0, inline = false}) {
+function toJSXString({ReactElement = null, lvl = 0, inline = false, options = {}}) {
   if (typeof ReactElement === 'string' || typeof ReactElement === 'number') {
     return ReactElement;
   } else if (!isElement(ReactElement)) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -145,28 +150,28 @@ function formatValue(value) {
     // we then remove the whole wrapping
     // otherwise, the element would be surrounded by quotes: <div a={{b: '<div />'}} />
     return '<__reactElementToJSXString__Wrapper__>' +
-      toJSXString({ReactElement: value, inline: true}) +
+      toJSXString({ReactElement: value, inline: true, options}) +
```
leoasis: as mentioned in the PR description, we need to pass around the `options` object in all these methods only to be able to pass it here (and below in the `recurse` function. I think it may be clearer to access it from the parent scope by moving all these functions inside the body of `toJSXString`

vvo: I think we only need to move `toJSXString` inside the export, that should be sufficient.

What do you think?

vvo:
Looks awesome, I added a small comment.

index.js
leoasis: we will also need to move `recurse` and `formatValue` so that they are able to call `toJSXString`, and then we will also need to move the functions that call those, and so on

vvo: > we will also need to move recurse and formatValue so that they are able to call toJSXString, and then we will also need to move the functions that call those, and so on

Ok just move everything inside the main export, so we only have to specify options once. works?

leoasis: Sorry for the delay, will do this as soon as I can!

leoasis:
@vvo I've pushed the changes we discussed, had to also rename the `ReactElement` variable in `toJSXString` to avoid shadowing.

vvo:
Thanks, merging this

vvo:
published as 2.2.0

leoasis:
Awesome, thanks for accepting the idea!

PR #19, opened 2016-03-08
Support React 14 and 15

ericf:
Adds support for React 15: http://facebook.github.io/react/blog/2016/03/07/react-v15-rc1.html


vvo:
Once its fully released, can we just say react: ^15.0.0?

ericf:
Yeah. Although I think this should work when 15.0.0 comes out too.

PR #23, opened 2016-04-08
Fix ES6 class component (ReactComponent.type not existing > using Rea…

totopsy:
…ctComponent.constructor) to get the name of the component

### DIFF
--- a/index.js
+++ b/index.js
@@ -164,6 +164,9 @@ export default function reactElementToJSXString(ReactElement, options = {}) {
 
 
 function getDefaultDisplayName(ReactElement) {
+  if (!ReactElement.type) {
+    return ReactElement.constructor ? ReactElement.constructor.name : 'No Display Name';
+  }
   return ReactElement.type.name || // function name
     ReactElement.type.displayName ||
     (typeof ReactElement.type === 'function' ? // function without a name, you should provide one


vvo:
Can you provide a test case because I am failing to see when can this break. Thanks

totopsy:
```
'use strict';

import React from 'react';
import { Link } from 'react-router';
import chai, {expect} from 'chai';
import jsxChai from 'jsx-chai';

chai.use(jsxChai);

//import {TransactionItem} from '../../../../src/js/components/transaction/transactionItem';

class TransactionItem extends React.Component{
    constructor(props){
        super(props);
    }
    componentWillReceiveProps(newProps){
    }
    render(){
        return (
            <div key={this.props.transaction._id} className={'listItem transactionListItem'}>
                <Link to={'/connections/' + this.props.transaction.connectionId + '/accounts/' + this.props.transaction.accountId + '/transactions/' + this.props.transaction._id + '/detail'}>
                    <div className='left content'>
                        <div className='category foodCategory'>
                            <span className='border'/>
                            <div className='food' />
                        </div>
                        <div className='itemTransactionLabel'>{this.props.transaction.simplifiedWording}</div>
                        <div className='itemTransactionValue'>{this.props.transaction.value}</div>
                    </div>
                    <div className={'icon openArrowIcon right'} />
                </Link>
            </div>
        );
    }
}

describe('TransactionItem', function() {
    const transaction = {
        "_id": 13627,
        "value": -34.75,
        "simplifiedWording": "SOCIETE GENERALE"};

    it ('checkSimplifiedWording', () => {
        expect(<TransactionItem transaction={transaction} />).to.include(transaction.simplifiedWording);
    })
    it ('checkWrongSimplifiedWording', () => {
        expect(<TransactionItem transaction={transaction} />).to.include('toto');
    })
});
```

vvo:
We have a full test suite: https://github.com/algolia/react-element-to-jsx-string/blob/master/index-test.js

Can you add a reduced test case in there?

PR #26, opened 2016-05-16
allow omitting boolean attributes

p0wl:
according to https://facebook.github.io/react/docs/jsx-in-depth.html#boolean-attributes it is ok to omit the value of boolean attributes if the value is true. Attributes with value false are unnecessary.

This PR allows (via options: {omitBools: true}) that

```
<Button primary={true}>A Button</Button>
```

becomes

```
<Button primary>A Button</Button>
```

I think this is a good idea in general, but I thought I'll create a option for this, as this is unexpected behavior for existing consumers. What do you think? Should the short syntax be the default?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Useful for unit testing and any other need you may think of.
 
 Features:
 - supports nesting and deep nesting like `<div a={{b: {c: {d: <div />}}}} />`
-- props: supports string, number, function (inlined as `prop={function noRefCheck() {}}`), object, ReactElement (inlined), regex..
+- props: supports string, number, function (inlined as `prop={function noRefCheck() {}}`), object, ReactElement (inlined), regex, booleans (with [shorthand syntax](https://facebook.github.io/react/docs/jsx-in-depth.html#boolean-attributes)), ...
 - order props alphabetically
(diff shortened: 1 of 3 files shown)


vvo:
Since it's not possible I guess to detect if input JSX was using short or long syntax, it would be easier to just transform to short syntax by default, this would be a breaking change but I am ok doing it.

What do you think?

vvo:
If you agree, then remove the option and I will bump major

p0wl:
I agree. I've removed the option commits and force pushed. documented the support of boolean attributes in the README.

If someone needs it as an option, the commit is still available here: https://github.com/Jimdo/react-element-to-jsx-string/tree/boolean-values-option

Would be great if you merge and bump. Thanks for your project! I'll use it in our pattern library soon!

vvo:
Awesome thanks for contributing, I was on vacation, hence the late reply. Now merging

vvo:
ok done let me know if it's ok

vvo:
where are you gonna use this module ?

p0wl:
sorry for the really late answer: we used it in our pattern-library to show how the examples look code wise, so the preview shows a button pattern and we use your awesome module to render `<Button primary />` as code.

PR #27, opened 2016-06-27
Remove default props

slightlytyler:
Removes default props if their value is the default. This should match JSX that was input (or at least it does in my scenario).

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -461,7 +461,7 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
 
   it('should not render default props if "showDefaultProps" option is false', () => {
     expect(
-      reactElementToJSXString(<DefaultPropsComponent />, { showDefaultProps: false })
+      reactElementToJSXString(<DefaultPropsComponent />, {showDefaultProps: false})
     ).toEqual(`<DefaultPropsComponent />`);
   });
 
--- a/index-test.js
+++ b/index-test.js
@@ -453,9 +453,15 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
     ).toEqual(`<TESTCOMPONENT />`);
   });
 
-  it('reactElementToJSXString(<DefaultPropsComponen />)', () => {
+  it('should render default props', () => {
     expect(
       reactElementToJSXString(<DefaultPropsComponent />)
+    ).toEqual(`<DefaultPropsComponent test="test" />`);
+  });
+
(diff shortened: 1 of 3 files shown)


index-test.js
```diff
@@ -449,6 +453,12 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
     ).toEqual(`<TESTCOMPONENT />`);
   });
 
+  it('reactElementToJSXString(<DefaultPropsComponen />)', () => {
+    expect(
+      reactElementToJSXString(<DefaultPropsComponent />)
```
vvo: > Removes default props if their value is the default.

I am not sure to follow what you are trying to prove in this test, shouldn't we use test='test' somewhere?

vvo:
Thanks for your contribution! I am looking forward at understanding the usecase.

index-test.js
```diff
@@ -449,6 +453,12 @@ describe(`reactElementToJSXString(ReactElement)`, () => {
...
+    ).toEqual(`<DefaultPropsComponent />`);
```
vvo: Shouldn't the output be DefaultPropsComponent test=test here?

slightlytyler: The point is that if a passed property is equivalent to the default it should not be shown. Maybe it should be an option than the default behavior. `<DefaultPropsComponent />` and `<DefaultPropsComponent test="test" />` are equivalent but the latter is showing props that are inferred from the defaults.

slightlytyler:
The use case for me is in my documentation components. I have an `Example` component that renders the children as well as renders their code in a block.
```javascript
<Example>
  <Button>Button</Button>
</Example>
```
screen shot 2016-06-27 at 9 39 52 am
Without my changes you see all the default props, which were not passed in but inferred, in the code block.

vvo:
Ok, can you create an option like "showDefaultProps", should be true by default if that was the current behavior.

slightlytyler:
@vvo should be good to go now

slightlytyler:
oops need to update docs actually

vvo:
awesome, I love the fact that you were able to hack this new option in the current code without my help :) GG!

vvo:
Where is the documentation you are working on?

slightlytyler:
@vvo it's easy to augment good code :wink: did you see the documentation? I added a blurb to the options section

vvo:
yes, perfect :)

PR #32, opened 2016-07-14
Add "showFunctions" option

TiuSh:
I found it useful to had an option to show functions' code in some cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,12 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   If false, default props are omitted unless they differ from from the default value.
 
+**options.showFunctions: boolean, default false**
+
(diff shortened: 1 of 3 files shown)


vvo:
That's nice! I wonder how we could keep the original function notation instead of the transpiled one, any idea?

TiuSh:
Good question! I'm looking into this...

vvo:
I am not sure this is easily doable, or even doable, in that case I am fine with the change

TiuSh:
I wondered if we could use Source Maps when generated by transpilers, but that's a dead end. Hopefully when es6 transpilers won't be needed anymore the output will respect the original notation :)

vvo:
Nice!

vvo:
@TiuSh published

PR #35, opened 2016-08-03
change order of where to get the name of a component from

msurekci:
Changed the order because `ReactElement.type.name` returns something that doesn't represent the name of the component.

For example if the component was called `Button` it would return a random letter instead like 'L' or 't'.

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -18,6 +18,10 @@ class DefaultPropsComponent extends React.Component {}
 
 DefaultPropsComponent.defaultProps = {test: 'test'};
 
+class DisplayNamePrecedence extends React.Component {}
+
+DisplayNamePrecedence.displayName = "This should take precedence";
+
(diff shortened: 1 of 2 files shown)


Morhaus:
Hey, thanks for the PR!

Indeed, `type.displayName` should take precedence over `type.name` when determining component names. Could you please write a test for this?

Also, in your particular case, I believe `ReactElement.type.name` returns a wrong result because the corresponding component's code is minified: `class MyComponent extends React.Component {}` => `class t extends React.Component {}`.

msurekci:
Thanks for the quick reply Morhaus. I've added the unit test.

Please let me know if you would like me to make any other changes.

msurekci:
@Morhaus when can I expect this change to be released on npm?

Morhaus:
@msurekci I've just pushed to npm 👍

PR #40, opened 2016-08-23
move babel-presets from devDep to deps

anback:
I have a react native application, i had to do this to get it to work, plz advice (and hopefully merge)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,6 @@
   "repository": "algolia/react-element-to-jsx-string",
   "devDependencies": {
     "babel-cli": "^6.10.1",
-    "babel-preset-es2015": "^6.9.0",
-    "babel-preset-react": "^6.11.1",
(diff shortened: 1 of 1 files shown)


vvo:
Hi @anback, I do not have experience with react-native but it seems not right to have to do so.

We publish an ES5 package (https://github.com/algolia/react-element-to-jsx-string/blob/2c8f8647b61ef17d28cea0cf13c41c58105038a8/package.json#L5-L7) so you should not have to have to compile this project yourself. Maybe we are missing something here.

The fact that .babelrc goes into the npm package is maybe the reason of your issues. Could you find maybe some reference/other projects having struggle with react native and babel as dev deps?

vvo:
On a side note I am curious to know how you are using this package with react-native?

vvo:
Closing, reopen if needed

PR #41, opened 2016-08-30
Added configurable option to explicitly output boolean prop values. 

HHogg:
For us weird folk that don't use shorthand syntax (I know I know) this adds a configurable option to output boolean values like:

```jsx
<Component prop={true} prop={false} />
```

Default is set to `true` for backwards compatibility ... and also because I imagine the majority use the shorthand syntax. 

_Also there were a couple of eslint errors, not sure if they were there before... I don't think I have any global eslint config madness sneaking in there_

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Useful for unit testing and any other need you may think of.
 
 Features:
 - supports nesting and deep nesting like `<div a={{b: {c: {d: <div />}}}} />`
-- props: supports string, number, function (inlined as `prop={function noRefCheck() {}}`), object, ReactElement (inlined), regex, booleans (with [shorthand syntax](https://facebook.github.io/react/docs/jsx-in-depth.html#boolean-attributes)), ...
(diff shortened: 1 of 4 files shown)


vvo:
> Also there were a couple of eslint errors, not sure if they were there before... I don't think I have any global eslint config madness sneaking in there

Ah yes sorry that's because we use Travis cache and then our eslint config was upgraded. Sorry about that

README.md
```diff
@@ -67,6 +67,12 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   Just return the name you want for the provided ReactElement, as a string.
 
+**options.showBooleans: boolean, default false**
```
vvo: Can we rename it to `useBooleanShorthandSyntax`? Long but explicit, showBooleans is not precise enough I believe

vvo:
Awesome! I have left one comment.

What do you use this library for?

README.md
HHogg: Yeah makes perfect sense I'll update to that :+1: with the default to `true`.

HHogg:
Thanks @vvo have updated with the rename 👍  Using this at the moment for a work in progress React pattern library, specifically for the style guide examples.

PR #43, opened 2016-09-20
`filterProps` Option

HHogg:
#### showKey
We have situations where we need to add a key property to the input of `reactElementToJsxString` but not to have it printed in the output. This adds a `showKey` option to not print the key prop.

#### lodash
When the import syntax like below is used
```
import {fill} from 'lodash';
``` 
it still requires the whole library, but changing it to 

```
import fill from 'lodash/fill';
```
It will only require that util.

Our build size before this change was `2,428,686 bytes` and afterwards it dropped to `1,898,514 bytes`

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   Just return the name you want for the provided ReactElement, as a string.
 
+**options.filterProps: array, default []**
+
+  Provide an array of props to filter for every component. For example ['key'] will suppress the key="" prop from being added.
+
 **options.showDefaultProps: boolean, default true**
 
   If true, default props shown.
--- a/index-test.js
+++ b/index-test.js
@@ -456,6 +456,14 @@ describe('reactElementToJSXString(ReactElement)', () => {
     ).toEqual('<TESTCOMPONENT />');
   });
 
+  it('reactElementToJSXString(<TestComponent />, { filterProps: [\'key\', \'className\'] })', () => {
+    expect(
+      reactElementToJSXString(<TestComponent a="b" key="aKey" className="aClass" />, {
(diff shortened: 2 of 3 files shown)


vvo:
Hi!

Can you split those two changes in two PR? I am unsure about the lodash import change. This to me looks like a build issue rather than an import style statement issue.

It should be feasible to only import fill while still keeping the nice import syntax. Have a look here: http://knpw.rs/blog/using-lodash

vvo:
Thanks for the PR! About showKey, maybe we could make it more generic and provide a props filtering method where you would get the propName and you should return true/false if you want to keep it?

That would make this new API entry more generic if people wants to say, remove className. What do you think?

HHogg:
Thanks @vvo both sound good! I was flip flopping between a generic prop filtering and a specific key option... "I chose poorly" 😉 will update 👍 

In that article you mentioned, it suggests using the babel plugin `babel-plugin-lodash` to handle the optimisations, and as most setups I've seen `node_modules` are excluded from going through babel so someone consuming the `react-element-to-jsx-string` library wouldn't be able to implement the optimisation themselves. This plugin would need to be added to the `react-element-to-jsx-string` build steps. 

Would you prefer this additional dependency to convert the import syntax over writing it manually?

vvo:
> This plugin would need to be added to the react-element-to-jsx-string build steps.
> 
> Would you prefer this additional dependency to convert the import syntax over writing it manually?

Yep this would be what we need to do. Only a matter modifying https://github.com/algolia/react-element-to-jsx-string/blob/master/.babelrc.

I do prefer this method mostly because I only use the import {form}, having to manually namespace my imports is not something we want to do.

Thanks a lot for your understanding :)

HHogg:
Ok cool :) no problem, will make those changes now.

vvo:
Released as 4.2.0. thanks

PR #44, opened 2016-09-21
Lodash optimisation

HHogg:
Lodash import optimisation using `babel-plugin-lodash` plugin to support `import {util} from 'lodash';` syntax.


vvo:
Thanks

PR #50, opened 2016-10-21
Add collapseWhitespace option that, when false, will render array and…

### LINKED ISSUE #49: Feature request: add an option to display plain object and array props without collapsing whitespace
Hey – love this utility! Use it all the time.

I've got some cases where certain props accept nested objects which can be difficult to read when they appear on a single line with whitespace removed. An example:

```jsx
<Menu items={[{id: 'one', label: 'One'}, {id: 'two', label: 'Two'}, {id: 'three', label: 'Three'}]} />
```

It'd be great to pass an option like `{ collapseWhitespace: false }` and see something like this instead:

```jsx
<Menu
  items={[
    {
      id: 'one',
      label: 'One'
    },
    {
      id: 'two',
      label: 'Two'
    },
    {
      id: 'three',
      label: 'Three'
    }
  ]}
 />
```

Not sure if that's of interest to you or anyone else, but on the off chance that it is I'll submit a PR momentarily.

beauroberts:
… plain object props across multiple lines and with proper indentation. Resolves #49


vvo:
Hi thanks again!

This more or less looks like a prettify option, don't you think we could make this the default. I am not sure there's a reason to not make it the default since it provides a very good value.

We could make it the default and forget about having an option if we make sure it's relevant in all cases, which seems to be the case.

beauroberts:
That sounds great! I wasn't sure if there was a reason for them to be collapsed onto a single line, which is why I went with the non-default option approach. I'll make that change before merging.

vvo:
Sorry you will have some conflicts, I just upgraded a bit eslint et al.

Also, how do you use this module? In tests? In documentation? Thanks

beauroberts:
No worries, @vvo! I made the changes you suggested. Would you like to have a look before I merge?

I'm using it for documentation! It's awesome.

vvo:
We are good to go, I will do the merge and release for this one, we need something in the CHANGELOG.

PR #68, opened 2016-12-07
Fix an issue where boolean props with a default value of `true` are not rendered when their value is `false`

### LINKED ISSUE #67: Boolean props with a default value of `true` are not rendered when the value is `false`
If you create a component with a default boolean prop set to `true`, like this:

```jsx
class DefaultPropsComponent extends React.Component {}

DefaultPropsComponent.defaultProps = {test: 'test', something: true};
```

and then render an instance of that component where the prop is `false`:

```jsx
reactElementToJSXString(<DefaultPropsComponent something={false} />, {showDefaultProps: false})
// Result: <DefaultPropsComponent />
```

the resulting string does not include `something={false}` as expected.

I'll submit a PR momentarily!

beauroberts:
Resolves #67


vvo:
Nice!

PR #70, opened 2016-12-08
Added .npmignore file

### LINKED ISSUE #69: Add .babelrc to .npmignore file
Hi I am using your library and for some reason my webpack build process is picking up your .babelrc file making my builds fail because I don't have all the plugins and presets you use in your library. This shouldn't be happening because my code is not building your library (and the plugins/presets from your library are under devDependencies), your library is compiled before publishing to `index-dist.js` so I shouldn't need any of your build-process stuff. If I remove `project_root/node_modules/react-element-to-jsx-string/.babelrc` file the build works as normal.

So all we need to do to fix these problems is adding a .npmignore file in the repository:
```
.babelrc
```

And republish the project to NPM.

Also thanks for the lib, pretty neat.

DanielHoffmann:
Removes configs and source code from the published bundle. Fixes #69. This PR is published to NPM under `react-element-to-jsx-string-test` if you want to test it

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+.babelrc
+.eslintignore
+.eslintrc.js
+.travis.yml
+*.js
+!index-dist.js
+mocha.opts


vvo:
GG!

PR #73, opened 2017-02-01
Symbol support fix #72

### LINKED ISSUE #72: Support Symbol prop types
As a developer I want to use Symbols as prop types

Given I have a component `<Button />`
And I want to give that component a type "Primary"
And I use `const TYPE_PRIMARY = Symbol('TYPE_PRIMARY')` to specify a button type
And I have the following react element `{ ...otherAttributes, props: { type: TYPE_PRIMARY } }`
And I pass  through react-element-to-jsx-string

Then I expect the resulting jsx string to be `<Button type={TYPE_PRIMARY} />`

But I get an error "Uncaught TypeError: Cannot convert a Symbol value to a string"

---

I understand that we do not have the actual variable name that is used, and we can not be sure that the symbols name is mapping to that.

I'd propose to add a way to extend `formatValue` via options, so that I as user of this lib, can decide that on my own.

smacker:
Fix #72

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -440,6 +440,12 @@ describe('reactElementToJSXString(ReactElement)', () => {
 </div>`);
   });
 
+  it('reactElementToJSXString(<div type={Symbol("test")}/>)', () => {
+    expect(
+      reactElementToJSXString(<div type={Symbol('test')}/>)
+    ).toEqual('<div type={Symbol(test)} />');
+  });
+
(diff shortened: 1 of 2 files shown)


vvo:
Nice! Are you working together with @Xiphe?

If not, @Xiphe does this PR solves your issue?

Xiphe:
Yeah! this should be just fine. <3

Xiphe:
@vvo could you release this on npm?

PR #74, opened 2017-02-22
Add maxInlineAttributes option

andrewiggins:
Would you be open to adding a `maxInlineAttributes` option (or something similarly named)?

I'm using your library to render demonstration documentation where I show a react component running on a page and the react component source on the same page. For some of my simpler components, I'd like to have all of their props render on the same line as the tag:
```jsx
<Header level={2} kind="heading1">
  Content
</Header>
```

See the tests I've added for more comprehensive examples of the behavior. All existing tests pass so this should not break any existing behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,11 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   If false, Boolean prop values will be explicitly output like `prop={true}` and `prop={false}`
 
-**options.maxInlineAttributes: number, default 1**
+**options.maxInlineAttributesLineLength: number, default undefined**
 
-  Provide the max number of attributes to render inline with the tag name. If the number of attributes exceeds this number, then all attributes will be rendered
-  on a separate line.
+  Provide the max number of characters the render inline with the tag name. If the number of characters exceeds this number, then all attributes will be rendered
+  on a separate line. The default behavior if this option is undefined is render attributes inline if there is only one, and render attributes on multiple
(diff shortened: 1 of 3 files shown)


vvo:
What about we create something based on the character length of the first line? Would that make more sense than the number of attributes on the first line?

vvo:
BTW thanks for opening this

andrewiggins:
Yup, that makes sense, I can do that.

vvo:
Ok great! Since most of the time we display code in a monospace font then having this parameter by the line length should be good yes.

Let me know when you're ready!

andrewiggins:
@vvo Take a look and let me know what you think.

Because we want to use the length of line if all attributes are rendered inline, I need to keep track of how long that line is. Once that line is longer than the user wants, I need to start rendering attributes as multi-line. However I can't change attributes that have previously been rendered. So instead, I loop through all the attributes and render them both as inline and multi-line. After both have been rendered I determine which one to use based on a variety of factors to keep backwards compatibility.

By default the new option is undefined, and everything renders as it did before this change (at least according to the tests). Once you give the new option a length, the new behavior kicks in.

Currently the new option is called `maxInlineAttributesLineLength` but I'm open to other names if you have better ideas.

andrewiggins:
Let me know if you'd like me to rebase my commits together or something like that

vvo:
Seems good 👌

schoenwaldnils:
@vvo could you publish this to npm? I can't wait to use this feature :D

vvo:
Yup

vvo:
done as 6.3.0

PR #75, opened 2017-03-03
Add functionValue option

giedrius-timinskis:


giedrius-timinskis:
Hey!

I really liked the way function output looked on [jsx-to-string](https://www.npmjs.com/package/jsx-to-string) and thought it'd be nice to take it one step further and allow the user to specify the value. Let me know what you guys think about it, and if you think it's a good idea please help me out with naming the option because I don't think functioValue is descriptive enough.

Cheers!

index-test.js
```diff
@@ -748,4 +748,16 @@ describe('reactElementToJSXString(ReactElement)', () => {
...
 </div>`);
   });
+
+  it('should return the functionValue value when "showFunctions" is true and functionValue is a valid string', () => {
+    expect(
+      reactElementToJSXString(<div fn={() => 'value'}/>, {showFunctions: true, functionValue: 'Replace me'})
+  ).toEqual('<div fn={Replace me} />');
+  });
+
+  it('should return noRefCheck when "showFunctions" is false and "functionValue" is provided', () => {
```
vvo: "and functionValue is not provided"

index-test.js
```diff
@@ -748,4 +748,16 @@ describe('reactElementToJSXString(ReactElement)', () => {
   </div>
 </div>`);
   });
+
+  it('should return the functionValue value when "showFunctions" is true and functionValue is a valid string', () => {
+    expect(
+      reactElementToJSXString(<div fn={() => 'value'}/>, {showFunctions: true, functionValue: 'Replace me'})
```
vvo: Can you make it so that functionValue is a function that should return a string and gets the original function code (fn.toString()) in input? I guess that'd be more generic from the ground up.

```js
functionValue: (originalCode) => {
  return 'whatever';
};
```

giedrius-timinskis: @vvo Could you elaborate on this more? What do you mean by `original function code` in this context? 

Thanks

vvo: when your component is used like this:

```jsx
reactElementToJSXString(<Component prop={i => i++} />, {functionValue: code => console.log(code)})

// outputs 'i => i++'
```

You can get this with fn.toString();

Example with node repl:

image

If this is too obscure, let me know and we ditch this.

giedrius-timinskis: @vvo I just tried it out and ran into an issue: 
`
it('should return the functionValue as a string when "showFunctions" is true and functionValue is a function', () => {
    expect(
      reactElementToJSXString(<div fn={() => 'value'}/>, {showFunctions: true, functionValue: i => i++})
  ).toEqual('<div fn={i => i++} />');
  });
`
By the time the arrow function reaches to `.toString`, it gets  transpiled to
`function functionValue(i) {
    return i++;
  }}`

Any advice on how I should go around this? Alternatively we could allow only a string.

vvo: > By the time the arrow function reaches .toString, it gets transpiled to
> function functionValue(i) { return i++; }}

No big deal, this is only because you must be running some babel transformation, when running in latest nodejs with no transforms, should work.

At least you get the right function body.

Let me know when it's ready.

giedrius-timinskis:
@vvo Let me know if the latest changes are what you had in mind. Thanks!

vvo:
The only API I would like to be done here is:

```jsx
reactElementToJSXString(<div fn={() => 'value'}/>, {showFunctions: true, functionValue: originalFunction => originalFunction.toString()})
```

Which means:

**functionValue** is a function (and only a function, not a string) that will be passed the originalFunction reference (not a string) and should return a string that will be used as a replacement for the originalFunction.

Sorry for the back and forth and for not being clear at first. This is the only API we need, it's pluggable, extensible. Thanks!

giedrius-timinskis:
@vvo Thanks for the feedback, and your patience :) . Please let me know if fa66240 is what you had in mind. 

I'm coercing the non-string result to a string, not sure if that's desired behaviour.

vvo:
Two small changes:
[ ] make it so that by default the code on functionValue will send "function noRefCheck() {}" string. So that this code: https://github.com/giedrius-timinskis/react-element-to-jsx-string/blob/fa6624088ebc5e6cbd3817ba87eb4b295b20ac0a/index.js#L196-L201 can be merged into a single functionality by definining the default functionValue implementation to return "function noRefCheck()"
[ ] update the documentation so that it reflects that the default functionValue is "function noRefCheck() {}"

We will upgrade the major version since I am unsure on the implications of the first [ ].

giedrius-timinskis:
@vvo Is dac0425 what you had in mind? Let me know, I'll need to squash the commits if we are to proceed.

vvo:
@giedrius-timinskis sorry it took some time but I did the changes myself see here: https://github.com/algolia/react-element-to-jsx-string/pull/75/commits/2885706eca82ab08aefdfde7fed0d214924e77f2

giedrius-timinskis:
Thanks @vvo

PR #79, opened 2017-05-23
Fix edge case when props contains double quote

shouze:
Looks like this case was not managed and that we encountered it 😉

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -57,6 +57,22 @@ describe('reactElementToJSXString(ReactElement)', () => {
     );
   });
 
+  it("reactElementToJSXString(React.createElement('div', {title: 'hello \"you\"'}))", () => {
+    expect(
+      reactElementToJSXString(
+        React.createElement('div', { title: 'hello "you"' })
+      )
+    ).toEqual('<div title="hello &quot;you&quot;" />');
+  });
+
+  it("reactElementToJSXString(React.createElement('div', {title: Symbol('hello \"you\"')})", () => {
+    expect(
+      reactElementToJSXString(
+        React.createElement('div', { title: Symbol('hello "you"') })
+      )
+    ).toEqual('<div title={Symbol(hello "you")} />');
+  });
+
   it('reactElementToJSXString(<div/>)', () => {
     expect(reactElementToJSXString(<div />)).toEqual('<div />');
   });
--- a/index-test.js
+++ b/index-test.js
@@ -65,6 +65,14 @@ describe('reactElementToJSXString(ReactElement)', () => {
     ).toEqual('<div title="hello &quot;you&quot;" />');
   });
 
+  it("reactElementToJSXString(React.createElement('div', {title: '<'hello' you & you>'}))", () => {
+    expect(
+      reactElementToJSXString(
+        React.createElement('div', { title: "<'hello' you & you>" })
+      )
+    ).toEqual('<div title="&lt;&#39;hello&#39; you &amp; you&gt;" />');
+  });
+
   it("reactElementToJSXString(React.createElement('div', {title: Symbol('hello \"you\"')})", () => {
     expect(
       reactElementToJSXString(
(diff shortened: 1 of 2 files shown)


vvo:
Hey hi nice! What was the result without your addition? Are there any other cases we should also escape?

index-test.js
```diff
@@ -57,6 +57,22 @@ describe('reactElementToJSXString(ReactElement)', () => {
     );
   });
 
+  it("reactElementToJSXString(React.createElement('div', {title: 'hello \"you\"'}))", () => {
+    expect(
+      reactElementToJSXString(
+        React.createElement('div', { title: 'hello "you"' })
+      )
+    ).toEqual('<div title="hello &quot;you&quot;" />');
```
shouze: Without that patch it was leading to this invalid JSX string:
```jsx
<div title="hello "you"" />
```

Haroenv:
Interesting, This case doesn’t come up with regular React.createElement, since it will show as `"hello 'you'"`, see https://codepen.io/Haroenv/pen/1de431341debbb029ee195fe8c24bc95/?editors=0011

shouze:
@Haroenv yes but the problem is that when you try to render `<div title="hello "you"" />` it would lead to an error, it's not about `React.createElement` https://codepen.io/shouze/pen/dWaeOd?editors=0012

Our error was this one, a babel jsx parse error (during a webpack build)


```
Module build failed: SyntaxError: Unexpected token (347:69)
```

shouze:
@vvo don't want to hurry you up but would be very pleased to see a patch release available today if LGTM 👼

vvo:
@shouze Sure that's doable, just from your POV and @Haroenv, any way we can do a generic escape much like the one React does rather than a replace on a specific char?

Are there any other chars that should be escaped no matter what?

vvo:
Also I am curious, how do you use this module? For docs, for tests?

Haroenv:
https://stackoverflow.com/a/7382028/3185307 is linked in the react docs, so I’d say, escape those too

see https://github.com/facebook/react/blob/b1b4a2fb252f26fe10d29ba60d85ff89a85ff3ec/src/renderers/dom/shared/escapeTextContentForBrowser.js

shouze:
We use this module to build static websites. We have several build step but basically here's what we do when we want to generate the app.js file:
- Convert some XML into react elements (plus some decorations).
- Convert react elements to JSXString with this module & create some readable stream.
- Bundle the stream through webpack.

Looks like a bit complicated but this way we can upgrade our component lib independently of our build service (or make our build service use any component lib version on the fly) & drive the entire website generation from a XML format very close to our JSX so expressing very well what we want to obtain as a final result.
And about the XML input, it's a constraint of our publication service as xsd validation is okay & we didn't find as good JSX validators as it exists good XML validators nowaday.

About the other chars I've checked as we've already encountered the case of `&` for example but it's ok we don't have edge cases with it. I would say it's even worst to escape / convert them into HTML entities as renderToString will lead to double / bad encoding the way they manage the entity conversion:
https://codepen.io/shouze/pen/jmdpJR?editors=0012

Otherwise I would have used [lodash escape as it's a dependency](https://lodash.com/docs/4.17.4#escape).

I suggest I can amend my commit with unit tests with `&`,  `<`, `>` & `'` chars to ensure we don't replace them by their entities and will not lead to some regression. WDYT?

shouze:
@Haroenv ok sorry I've double checked in fact: https://codepen.io/shouze/pen/jmdpJR?editors=0012

So I could use `lodash escape`, will do some extra processing but probably more secure.

shouze:
@vvo @Haroenv I've pushed a commit using lodash escape. Can squash it you're ok.

index.js
```diff
@@ -6,6 +6,7 @@ import stringify from 'stringify-object';
 import sortobject from 'sortobject';
 import traverse from 'traverse';
 import { fill } from 'lodash';
+import { escape } from 'lodash';
```
Haroenv: can me merged with line 8:

```js
import { fill, escape } from 'lodash';
```

shouze:
@Haroenv ok so it's fixed about the import 👼

vvo:
Will bump major as this is a breaking change, the output format of previous elements changed (&, ..) so people using this for tests would see a failure now. No big deal this is just to explain why we will do that

PR #81, opened 2017-05-26
Add failing test for unnecessary string escaping

beauroberts:
Occurs in nested JSX that contains single quotes

Issue #80

PR #82, opened 2017-05-30
Fix unnecessary string escaping

shouze:
Rel to #81


vvo:
I am kind of lost in my own code. Unable to see any shortcomings from this fix.

What about components that will have text containing ', " or \' as text?

PR #84, opened 2017-06-09
Remove props sort as it leads to isomorphism issue

shouze:
I haven't fixed the tests yet as I want to discuss 1st about the
(probably good) reason of removing this props sort or make it an option
to toggle.

In our use case of this lib, this leads to invalid checksums for
universal/isomorphic rendering.

### DIFF
--- a/index.js
+++ b/index.js
@@ -185,9 +185,9 @@ got \`${typeof Element}\``
       );
     }
 
-    return formatted
-      .sort()
-      .map(propName => getJSXAttribute(propName, props[propName], inline, lvl));
+    return formatted.map(propName =>
+      getJSXAttribute(propName, props[propName], inline, lvl)
+    );
   }
 
   function getJSXAttribute(name, value, inline, lvl) {


shouze:
@vvo so tell me what do you prefer, do I fix the tests or do I make this props sort an option?

shouze:
So far inside our own project we've added a sort on our component's props to fix this isomorphic issue so there's no emergency about this PR.

vvo:
Can you give me more context about what is the current issue for you? Go as detailed as you like, I am dumb! :)

Spy-Seth:
From a dynamically generated react elements tree, we build some static HTML (for isomorphic rendering) and (with the help of this lib) a compiled JS bundle (that we be mounted on the HTML on client side). 

Our issue is that when we try to mount our bundle on the HTML, React could not reuse the existing HTML markup because of some attribute on HTML tag that are not in the same order. As react compare strings the order of the attribute matter.

To generate the JS bundle we use `react-element-to-jsx-string` to help us to generate some JSX file. And this JSX file will be passed to bundler (webpack) to generate an usable bundle on client side. The issue is that during the JSX file, `react-element-to-jsx-string` does not preserve the attribute order on native HTML tags (on component tags the props order is not important). So when the bundled code is executed in client side we cannot correctly mount it on the HTML.

Two solve this issue we see two possibles solutions:
- Add an option to not sort the attribute/props
- Detect if the current component is a native HTML and disable the props ordering

What did you think?

vvo:
> Add an option to not sort the attribute/props

Yes for this, with a simple test. The sort was actually made default because this library was never designed to do what you are doing (but I love it). It was designed with string comparisons and diffs in mind (thus, we wanted "sort object keys in a deterministic order").

Add a options.sortAttributes (default true) and a test then we will be fine.

PR #86, opened 2017-06-11
always filter out props that have a value of undefined

arahansen:
The logic for filtering out defaultProps wasn't taking into account falsey values that are actually defined as default props.

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -16,7 +16,12 @@ function NamedStatelessComponent(props) {
 
 class DefaultPropsComponent extends React.Component {}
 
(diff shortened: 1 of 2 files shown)


vvo:
Thanks a lot, nice fix. What do you use this module for?

arahansen:
No problem! I'm using a storybook plugin, storybook-addon-jsx that uses this library under the hood.

arahansen:
I would love to use update that addon with my changes here. What would it take to get a release cut to npm? Thanks!

vvo:
10.0.1 now available

Issue #80, opened 2017-05-26
React elements nested in array/object props contain unnecessary string escaping


beauroberts:
Hi - thanks for an awesome tool!

I think this might be a regression but I'm not sure when it would have been introduced. I'm having issues with unnecessary string escaping on props that contain React elements with single quotes (e.g. `<div object={{ nested: <div array={['hello', 'you']} /> }} />`

I'll submit a pull request with a failing test but unfortunately I'm not able to investigate a fix right now. If no one else can look into it, I might get around to it in a few days:

```diff
+ expected - actual

  <div
    obj={{
-     nested: <div arr={[\'hello\', \'you\']} />
+     nested: <div arr={['hello', 'you']} />
    }}
   />
```

vvo:
cc @shouze maybe?

vvo:
might be a regression yes

PR #91, opened 2017-06-27
Add failing test on JSON escaping issue

Spy-Seth:
This test is here to discuss about the change made by the commit https://github.com/algolia/react-element-to-jsx-string/commit/5c0ae3ef718259e23510106bc4d0093dd88085d5 to solve the issue #80 on string escaping. The proposed fix introduce a regression on a the JSON representation that is no more correctly escaped.

@vvo I would have your thought about this one, I will propose a PR to correctly fix both cases

### DIFF
--- a/index-test.js
+++ b/index-test.js
@@ -92,6 +92,26 @@ describe('reactElementToJSXString(ReactElement)', () => {
     );
   });
 
+  it("reactElementToJSXString(<div nested={{ hello: 'world', foo: 'esca'ped', bar: 42 }} root=\"root\"/>)", () => {
+    expect(
+      reactElementToJSXString(
+        <div
+          nested={{ hello: 'world', foo: 'esca\'"ped', bar: 42 }}
+          root="root"
+        />
+      )
+    ).toEqual(
+      `<div
+  nested={{
+    bar: 42,
+    foo: "esca'\\"ped",
+    hello: "world"
+  }}
+  root="root"
+/>`
+    );
+  });
+
   it("reactElementToJSXString(React.createElement('div', {title: Symbol('hello \"you\"')})", () => {
     expect(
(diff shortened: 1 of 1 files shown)


vvo:
Hello, can you provide more context/use case/examples on why the current library state is not good? Sorry there have been a lot of small fixes to that output recently, it's hard to follow.

Phrase it as if I was not the maintainer :D

Spy-Seth:
Lately there have been a lot of change on the simple/double quote representation. The aim of all this modification is to always generate a valid output (correctly escaped).

The test that I've added show a last issue with the escaping:
screenshot 2017-06-28 17 25 41

After some digging the nested string in object props have simple quote as delimiter instead of double quote. All the string escaping system is made to generate string valid in double quote not in single quote. So actually single quote are not escaped witch led to erroneous string representation `'foo'bar'`. 

A first solution is to force all the delimiter to be double quote: `"foo'bar"`. This could be done by adding the options `{ singleQuotes: false }` to the `stringify` call.

A second solution is to keep the revert the https://github.com/algolia/react-element-to-jsx-string/commit/5c0ae3ef718259e23510106bc4d0093dd88085d5 commit who introduce the current issue. Previous this commit simple quote where correctly escaped. But we will to find another solution for quote in case of array representation: `[\'foo\']`

vvo:
I think this module should use double quotes by default for arrays and strings, that would solve this issue right?

Is it what you mean by using singleQuotes: false?

If so, let's do this.

vvo:
I think this module should use double quotes by default for arrays and strings, that would solve this issue right?

Is it what you mean by using singleQuotes: false?

If so, let's do this.

Spy-Seth:
Fixed by https://github.com/algolia/react-element-to-jsx-string/pull/107

PR #106, opened 2017-07-07
refactor: Fix eslint errors

rarkins:
This PR fixes the new eslint errors found in the existing Pin Dependencies PR


vvo:
As discussed via email, let's wait for the current rewrite.

PR #107, opened 2017-07-07
Rewrote parsing

Spy-Seth:
After some work, I've rewrote the core to fix the PR #91. 

Don't be scare by the diff size 😅  It's mainly copy/paste of the existing code into more isolated/dedicated sub module. I split the application in two step: parsing and formatting. 

I've upgrade/cleanup the dependencies and replace mocha by jest.

~~The test suite have 4 failures related to function formatting in the output. 3 of them appear after deps upgrade so I'm not worried.~ I've not changed the existing test, so everything should be backward compatible 🎉 

I will publish a beta version of my fork and use it on our application to check if everything is ok. 

I've still some works to do: fix last lint issues, fix flow issues, fix some FIXMEs. I will go into it later next week :)

I hope this PR do not introduce too many change in the code ^^ I could understand if you do not want to merge it back

# Changelog

## Summary

This major release do not introduce any new feature and should not introduce regressions. It only fix some issue about escaping in nested component.

## Internal change details

- Heavy internal rewrite of the library
  - Split in multiple files
  - Separate in two clear step: parsing and formatting
  - Add some unit tests
- Dependencies upgrade and cleanup
- Replace Mocha by Jest as test runner

# TODO
- [x] Fix main tests suite `src/index.spec.js`
- [x] Add more unit tests on each module
- [x] Fix lint issues
- [x] ~Fix flow issues~ Too many unresolvable flow error: we will not add it to the CI for now
- [x] Write a changelog
- [x] Edit package.json to publish again on algolia organisation
- [ ] Prepare next major version


vvo:
As for the summary this seems completely legit so I encourage you moving forward. Ideally having a second owner of this whole code is perfect for the library so I am glad you are doing this. 

Dont be affraid by renovate PRs Ill resetup it afterwards.

Ill give a more complete review laters this weekend

vvo:
Also splitting in different files is a great idea

src/formatter/formatComplexDataStructure.js
```diff
@@ -0,0 +1,67 @@
+/* @flow */
+
+import collapse from 'collapse-white-space';
```
vvo: What do you mean by "complexDataStructure"? Seems to be objects or arrays that could also be

src/formatter/formatProp.js
```diff
@@ -0,0 +1,57 @@
...
+import type { Options } from './../options';
+
+export default (
+  propName: string,
+  hasValue: boolean,
+  value: any,
+  hasDefaultValue: boolean,
+  defaultValue: any,
+  inline: boolean,
+  lvl: number,
```
vvo: What does lvl means?

src/formatter/formatProp.js
```diff
@@ -0,0 +1,57 @@
+/* @flow */
+
+import spacer from './spacer';
+import formatPropValue from './formatPropValue';
+import type { Options } from './../options';
+
+export default (
+  propName: string,
+  hasValue: boolean,
+  value: any,
```
vvo: we should either use name and value or propName and propValue

src/formatter/formatComplexDataStructure.js
```diff
@@ -0,0 +1,67 @@
...
+import { isValidElement } from 'react';
+import stringify from 'stringify-object';
+import sortobject from 'sortobject';
+import traverse from 'traverse';
+import parseReactElement from './../parser/parseReactElement';
+// import formatPropValue from './formatPropValue';
```
vvo: miss?

vvo:
I had a look, what I miss to better undertand:
- are we providing any new feature?
- are we incorporating breaking changes?
- what is the general idea of the program now? Last one what mostly a recursive loop parsing done the tree as a single file, can you outline the general idea and responsibilities? Going over some files it seems that they are sharing many aspects sometime (formatProp requires parse and also formatters). What may help is go even further and try to reduce interdependencies and only plug everything together inside src/index.js (this may be very vague)
- did you ever looked at how jest does it (snapshots of react elements?) might have some good architecture ideas and maybe things to incorporate.

Overall feeling is that the codebase is more complex BUT that's only because you have spent a good amount of time rebuilding it. So I trust you for the future of the lib.

Let me know when you are ready for another review.

Spy-Seth:
> - are we providing any new feature?
> - are we incorporating breaking changes?

No! The aim of this PR is to keep strict feature parity. The only change introduce is to try definitely fix the escaping issue. One other goal is to make the library more robust by making some implicit explicit.

> - what is the general idea of the program now? Last one what mostly a recursive loop parsing done the tree as a single file, can you outline the general idea and responsibilities?

This has been separate in two part: 
- one recursive loop that convert the react element into a normalized tree
- another loop that format the tree into a string by using specialized formatter 

The main change is that I remove some implicit context in the formatter to simplifies them and make their result more predictable. Each formatter now only manage the thing they should now. If something is transverse between multiple formatter (witch is rare), it is explicitly pass as an argument.
For example [the hack that help manage component in props](https://github.com/algolia/react-element-to-jsx-string/blob/master/index.js#L207): `__reactElementToJSXString__Wrapper__`. This hack is needed because in some case we have some children that have (maybe) be converted into string that need escaping. 

> Going over some files it seems that they are sharing many aspects sometime (formatProp requires parse and also formatters). 

This is the tradeoff of the change previously explained (removing context and implicit). 

One solution to avoid the use parser inside the formatter could be to increase the responsibility of the initial parse time. Actually the parser do not explore the props (it only copy them). The props explorations are done during the formatting phase. If we change this to explore the props in the parser, we will remove the need to call parser in the formatted. But we will need to add lots of complexity into the parser and in the tree format to manage all the use case. I'm not sure it worth it. 

>  What may help is go even further and try to reduce interdependencies and only plug everything together inside src/index.js (this may be very vague)

I admit this was my initial goal. But the complexity of the parser that need to be done have stop me (see previous paragraph)

> did you ever looked at how jest does it (snapshots of react elements?) might have some good architecture ideas and maybe things to incorporate

I have never looked the implementation of the jest `pretty-format` package. I will have a look into it. You right it could have some good implementation idea 👍

src/formatter/formatComplexDataStructure.js
Spy-Seth: Yes `complexDataStrucure` is for Object and Array. Things that need exploration to be converted. After reading the jest core, [they use `printComplexValue` naming](https://github.com/facebook/jest/blob/master/packages/pretty-format/src/index.js#L878) for this use case

src/formatter/formatComplexDataStructure.js
Spy-Seth: Oups!

src/formatter/formatProp.js
Spy-Seth: The `lvl` is the current formatting depth. Could be interpreted as the current indentation level. I do not change this parameter behavior. It could be rename to something more obvious

src/formatter/formatProp.js
Spy-Seth: Good catch!

vvo:
> One solution to avoid the use parser inside the formatter could be to increase the responsibility of the initial parse time

Do you mean having an index.js that would be more explicit about the program flow? Might be good but if that's not a path you feel like could make the program better. I'll understand

vvo:
Overall I am really happy with your direction and as soon as you reach a state that you think it's very good (not just solving your own issue). Then we can speak about merging.

It would also make you the main knowledgable person of the code, which means that people will expect you to be responsive to issues and features. I will also look at them and provide guidance in API and code but still, you may be the master know if we merge this and it involves some responsibilities.

On the API side, if you see anything we could ditch or some deprecation we could remove and infer a new major release. Now might be the good time too.

Spy-Seth:
> Do you mean having an index.js that would be more explicit about the program flow

You means something more precise than the actual `format(parse(element))`?

> you may be the master know if we merge this and it involves some responsibilities

I hope the change that I introduce make the code more readable for everyone and make external contribution easier 😄 

> On the API side, if you see anything we could ditch or some deprecation we could remove and infer a new major release. Now might be the good time too

The API looks good to me for now. The main change I will want to introduce at some point will be rework the test suite avoid duplicate test case and make it more organized: separate the integration tests (`src/index.spec.js`) from the unit tests, better organization of the test (especially for the related to an options). This should be done in another PR I think. This PR introduce to many change in the code base to allow me to change the test suite with confidence.

Spy-Seth:
Sorry for the delay, I could work again on this PR this week :) 

@vvo I think we are close to think about merging this. Did I miss some change you think mandatory?

Spy-Seth:
I have test [the last beta version published](https://www.npmjs.com/package/@rezzza/react-element-to-jsx-string) with one of out app that use `react-element-to-jsx-string` to convert (big and complex: 5k loc after output) react element: it's works well (and fix the original escaping issue). For me it's a good signal of merge ;)

vvo:
Tests are failing

vvo:
Will merge this today

vvo:
This is mostly gonna fail everytime because travis timezone is not UTC it seems so the output will always be CET.

Spy-Seth:
Yes! This show us we do not manage the date correctly. I manage this case :)

src/formatter/formatPropValue.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    );
+
+    expect(parseReactElement).toHaveBeenCalledTimes(1);
+    expect(formatTreeNode).toHaveBeenCalledTimes(1);
+  });
+
+  it('should format a date prop value', () => {
+    expect(
+      formatPropValue(new Date('2017-01-01T11:00:00.000Z'), false, 0, {})
+    ).toBe('{new Date("2017-01-01T11:00:00.000Z")}');
```
vvo: This is a "breaking change" right?

Spy-Seth: Yes I guess this is one. 

The escaping of the brace in string could be considered as breaking too (see f14ba62)

vvo: You can consider any change in the output of this library as a breaking change. Could you list them?

Spy-Seth: I think there are three breaking in the output:

- Fix escaping issue with quotes (in props value or in children of type string)
- Handle props value of type `Date`: `<div foo={new Date("2017-01-01T11:00:00.000Z")} />`
- Escape brace chars (`{}`) in a children string:  
```jsx
// The JSX:
<script type="application/json+ld">{`{ hello: 'world' }`}</script>
// Will become:
<script type="application/json+ld">&lbrace; hello: 'world' &rbrace;</script>

Comment (vvo) [inline src/formatter/formatPropValue.spec.js]: This PR is good to be merged, can you merge it in a single commit (just use the squash and merge button), inside the commit should be something like:

```text
fix(escaping): Complete lib refactor to handle more escaping cases

This is a complete refactor of the lib with:
- splitting into different modules..
- flow
- simplifying ..

BREAKING CHANGES:
- list any output breaking change
```

This is an example but you get the idea.

Then you can release the project (npm run release major). You have all the necessary rights. As I said, you should now be the main maintainer/knowledgeable person :)

Spy-Seth: > This PR is good to be merged

PR #111, opened 2017-07-21
Fix flow errors

Spy-Seth:


Spy-Seth:
Only one flow error left! The other is a warning.

Spy-Seth:
It's working!

PR #115, opened 2017-07-24
fix(flow): export flow definitions in the released bundle and fix the npm ignore too restrictive

Spy-Seth:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Turn a ReactElement into the corresponding JSX string.",
   "main": "dist/index.js",
   "scripts": {
-    "build": "babel ./src/ --ignore=*.spec.js --source-maps --out-dir ./dist/",
+    "build": "babel ./src/ --ignore=*.spec.js --source-maps --out-dir ./dist/ && flow-copy-source -v --ignore=*.spec.js src/ dist/",
```
Spy-Seth: Flow have an experimental command `flow gen-flow-files` to generate each file definition, but it's really experimental and failed too many times. So for now, we use the `flow-copy-source` tricks to have the flow definition exported with the library

PR #118, opened 2017-07-24
fix(release): revert change made by error in commit 86697517

Spy-Seth:


vvo:
Wonder if we could detect that we need to be run with npm run release or yarn release in a nice way. I guess not. Or we would have to detect if those binaries are available :D

Anyway thanks for the update

PR #119, opened 2017-07-25
Configure Renovate to rebase stale PRs

rarkins:
This repository appears to have a setting that all PRs must be up-to-date with the base branch before merging. For example, #117:

PR#117

Therefore it is recommended to enable the setting `rebaseStalePrs` in Renovate so that it knows to keep branches up-to-date itself and save you the task of doing it. It means Renovate will rebase any of its PRs off master any time it runs and finds that the master branch has changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
   },
   "license": "MIT",
   "repository": "algolia/react-element-to-jsx-string",
+  "renovate": {
+    "rebaseStalePrs": true
+  },
   "devDependencies": {
     "babel-cli": "6.24.1",
     "babel-eslint": "7.2.3",


vvo:
Thanks a lot, should this be the default of renovate?

rarkins:
> should this be the default of renovate?

I think that by default, Renovate should detect this setting at Onboarding time and recommend the setting. 

I'm still considering whether or not to make it default behaviour regardless though, because it will result in an increased number of builds and some people may not like that and instead prefer to rebase PRs themselves once they're ready to review.

vvo:
Would say for public projects, by default, people would be very happy if the PR got updated by themselves, less manual work.

And maybe in some rare cases, people will want to do it manually. But from my own experience, having to manually click on "update this branch" is painful.

PR #120, opened 2017-07-25
chore: Use semantic commit messages for Renovate

rarkins:


vvo:
Your view on this setting again, should it be the default? I does not harm projects that are not conventional changelog based, while it does provide values for the ones that are based on it.

rarkins:
@vvo the onboarding process detects if a repo uses semantic commit messages already, and configures it to true if so. So for projects already using semantic commits, they will get them - and for those not using semantic commits, they will have cleaner, shorter commit messages.

In the case of this repo, it was detected and recommended to use semantic commit messages: https://github.com/algolia/react-element-to-jsx-string/blob/9807dd54b0d3a35be09e3949cb9b2f5cf1d92084/renovate.json#L14

However as the PR was closed unmerged (no `renovate.json` added to `master`), that meant the repo reverted back to default settings.

rarkins:
By the way, I have changed the onboarding approach a little since you enabled this repository. Now when repositories are onboarded, the `renovate.json` only includes non-default settings so that it's more obvious to people like yourself about which settings are actually important/non-default.

vvo:
Thx!

rarkins:
FYI I've decided to go with your recommendation and rebase automatically by default if a repository is set to enforce "up-to-date" branches, assuming GitHub has an API that lets me detect that. See https://github.com/singapore/renovate/issues/524 for issue in progress

vvo:
Thanks a lot @rarkins we really appreciate what you do for renovate. I have been trying to push it hard on hackernews without success ahah but we'll get there. This is a very good piece of open-source!

rarkins:
@vvo thanks for your support! I have an idea for the next version which I think is a "killer" and maybe after that we push it again :)

PR #125, opened 2017-08-01
Allow react 16 as a peer dependency

gpeal:
Ran tests with react 15 and react 16 and all passed.
This is necessary to use with react native >= 0.43 (latest is 0.47-rc).
@Spy-Seth @vvo

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "react-test-renderer": "15.6.1"
   },
   "peerDependencies": {
-    "react": "^0.14.8 || ^15.0.1"
+    "react": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^16.0.0-beta || ^16.0.0-alpha"
   },
   "dependencies": {
     "collapse-white-space": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "react-test-renderer": "15.6.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
     "react-test-renderer": "15.6.1"
   },
   "peerDependencies": {
-    "react": "^0.14.8 || ^15.0.1"
+    "react": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^16.0.0-beta || ^16.0.0-alpha"
```
Spy-Seth: Could you also allow the release candidate?

Spy-Seth:
#127 is green! 🎉

gpeal:
@Spy-Seth They're green :)

package.json
Spy-Seth: This is still missing :)

gpeal:
@Spy-Seth Done!

vvo:
Hey hi @gpeal, thanks for the contribution. How are you using this library today? What's your usecase?

gpeal:
We (Airbnb) use chai-enzyme (which has a [dep](https://github.com/producthunt/chai-enzyme/blob/master/package.json#L88) of react-element-to-jsx-string) and are upgrading to React Native 0.46 which has a dep of react-16.0.0-alpha.12 so we need to go and update all of our deps and their transitive deps.

Issue #128, opened 2017-08-02
Don't mess json & jsx syntaxes - don't escape json in formatTreeNode


shouze:
In [formatTreeNode()](https://github.com/algolia/react-element-to-jsx-string/blob/34c395261d88771449d226af92b0ff5f45fd1b22/src/formatter/formatTreeNode.js#L16-L18) we escape strings & numbers without going further in the kind of what the string represents. We recently had the case of a ld+json string being inside a script tag:
```html
<script type="application/ld+json">
{
    "@context": "http://schema.org",
    "@type": "Hotel",
    "name": "Le Figuier",
    "description": "L’Hôtel Le Figuier est un petit établissement confortable et familial, au coeur d’Ivry-sur-Seine, à seulement 2km de Paris.",
    "address": "72 Boulevard de Brandebourg, 94200 Ivry-sur-Seine, France",
    "telephone": "+33 1 46 72 72 72",
    "photo": "https://res.cloudinary.com/karibbu/image/upload/c_scale,w_1800/v1501161224/lefiguier94.fr/VLWPuJuF7uUKLwMKZTytxg.jpg",
    "image": "https://res.cloudinary.com/karibbu/image/upload/s--P27nN2Ub--/v1501162649/lefiguier94.fr/Logo-lefiguier-light-bkg.svg",
    "starRating": {
        "@type": "Rating",
        "ratingValue": 0
    }
}
</script>
```

Yes it's a string, but it's a pure json string, encapsulated into script tags. So ATM the behavior escape it so it's rendered like that:
```html
<script type="application/ld+json">
&lbrace;
    "@context": "http://schema.org",
    "@type": "Hotel",
    "name": "Le Figuier",
    "description": "L’Hôtel Le Figuier est un petit établissement confortable et familial, au coeur d’Ivry-sur-Seine, à seulement 2km de Paris.",
    "address": "72 Boulevard de Brandebourg, 94200 Ivry-sur-Seine, France",
    "telephone": "+33 1 46 72 72 72",
    "photo": "https://res.cloudinary.com/karibbu/image/upload/c_scale,w_1800/v1501161224/lefiguier94.fr/VLWPuJuF7uUKLwMKZTytxg.jpg",
    "image": "https://res.cloudinary.com/karibbu/image/upload/s--P27nN2Ub--/v1501162649/lefiguier94.fr/Logo-lefiguier-light-bkg.svg",
    "starRating": &lbrace;
        "@type": "Rating",
[...]

Issue #126, opened 2017-08-02
Escape strings of HTML within <code> blocks


davidtheclark:
I have an element that looks like this:

```js
{ 
  '$$typeof': Symbol(react.element),
  type: 'code',
  key: 'h-4',
  ref: null,
  props:
   { className: 'language-javascript',
     children: [ '<div style={{ color: pink }}>123459254321</div>\n' ] },
  _owner: null,
  _store: {} 
}
```

It's a `<code>` element with JSX inside of it. 

The output from this module gives no special treatment to the JSX within the `<code>` element, so it ends up coming out as regular JSX: `<code><div style=...></div></code>`. Re-render that JSX and you have something different than what you started with.

I think this situation could be represented in JSX like `<code>{'<div style=...'}</div>`. You could also escape the `<` `>` signs, maybe.

I also think the same problem would apply to HTML within `<code>` blocks.

Do you think I'm misunderstanding something, or is this a bug?

Spy-Seth:
Could you provide of the JSX that will led to this  example of react element?

davidtheclark:
@Spy-Seth: Here's some JSX that will produce the problem:

```js
<code>
  {'<div style={{ color: pink }}>123459254321</div>\n'}
</code>
```

That renders to the element object I pasted above. When you feed that element through this module, though, you come out with the following *very different* JSX:

```js
<code>
  <div style={{ color: pink }}>123459254321</div>

</code>
```

davidtheclark:
@Spy-Seth that sounds like an excellent plan 👍

PR #130, opened 2017-08-02
[DO NOT MERGE] Detect stringified json children and escape them correctly

Spy-Seth:
Fix https://github.com/algolia/react-element-to-jsx-string/issues/128

The original aim of this PR was to manage JSON child. It also fix a bug with multi-line children string that were not correctly indented.


Spy-Seth:
After some second thought, it seems there a better way to solve the issue. See the issue https://github.com/algolia/react-element-to-jsx-string/issues/126#issuecomment-319675242 who talk about a similar issue.

Spy-Seth:
This PR is replaced by https://github.com/algolia/react-element-to-jsx-string/pull/133

PR #133, opened 2017-08-03
Fix string formatting edge case

### LINKED ISSUE #126: Escape strings of HTML within <code> blocks

### LINKED ISSUE #128: Don't mess json & jsx syntaxes - don't escape json in formatTreeNode

Spy-Seth:
Fix #126
Fix #128

This will lead to a major release

### DIFF
--- a/src/formatter/formatTreeNode.js
+++ b/src/formatter/formatTreeNode.js
@@ -4,8 +4,17 @@ import formatReactElementNode from './formatReactElementNode';
 import type { Options } from './../options';
 import type { TreeNode } from './../tree';
 
-const escape = (s: string): string =>
-  s.replace(/{/g, '&lbrace;').replace(/}/g, '&rbrace;');
+const jsxStopChars = ['<', '>', '{', '}'];
+const shouldBeEscaped = (s: string) =>
+  jsxStopChars.some(jsxStopChar => s.includes(jsxStopChar));
+
+const escape = (s: string) => {
+  if (!shouldBeEscaped(s)) {
+    return s;
+  }
+
+  return `{\`${s}\`}`;
+};
 
 export default (
   node: TreeNode,
@@ -13,8 +22,12 @@ export default (
   lvl: number,
   options: Options
 ): string => {
-  if (node.type === 'string' || node.type === 'number') {
-    return node.value ? escape(node.value.toString()) : '';
+  if (node.type === 'number') {
+    return String(node.value);
+  }
+
+  if (node.type === 'string') {
+    return node.value ? escape(String(node.value)) : '';
   }
 
   if (node.type === 'ReactElement') {
--- a/src/formatter/formatReactElementNode.js
+++ b/src/formatter/formatReactElementNode.js
@@ -8,8 +8,43 @@ import propNameSorter from './propNameSorter';
 import type { Options } from './../options';
 import type { ReactElementTreeNode } from './../tree';
 
-const recurse = (lvl: number, inline: boolean, options: Options) => element =>
-  formatTreeNode(element, inline, lvl, options);
+const compensateMultilineStringElementIndentation = (
+  element,
+  formattedElement: string,
+  inline: boolean,
+  lvl: number,
+  options: Options
+) => {
+  const { tabStop } = options;
+
+  if (element.type === 'string') {
+    return formattedElement
+      .split('\n')
+      .map((line, offset) => {
+        if (offset === 0) {
+          return line;
+        }
+
+        return `${spacer(lvl, tabStop)}${line}`;
+      })
+      .join('\n');
+  }
+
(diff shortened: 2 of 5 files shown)


src/formatter/formatTreeNode.js
```diff
@@ -4,17 +4,30 @@ import formatReactElementNode from './formatReactElementNode';
 import type { Options } from './../options';
 import type { TreeNode } from './../tree';
 
-const escape = (s: string): string =>
-  s.replace(/{/g, '&lbrace;').replace(/}/g, '&rbrace;');
+const jsxStopChars = ['<', '>', '{', '}'];
+const shouldBeEscaped = (s: string) =>
+  jsxStopChars.some(jsxStopChar => s.includes(jsxStopChar));
+
+const escape = (s: string) => {
```
shouze: 👍

PR #172, opened 2017-09-15
fix(formatting) Escape multi lines string nodes

shouze:
I consider that like a fix as for example we have the case of some markdown content string node:
if it's not protected then the carriage returns are lost at some point later when you render it.

We can also discuss that as a feature trough a toggled option.

### DIFF
--- a/src/formatter/formatReactElementNode.spec.js
+++ b/src/formatter/formatReactElementNode.spec.js
@@ -111,17 +111,17 @@ describe('formatReactElementNode', () => {
 
     expect(formatReactElementNode(tree, false, 0, defaultOptions)).toEqual(
       `<div>
-  first line
+  {\`first line
   second line
-  third line
+  third line\`}
 </div>`
     );
 
     expect(formatReactElementNode(tree, false, 2, defaultOptions)).toEqual(
       `<div>
-      first line
+      {\`first line
       second line
-      third line
+      third line\`}
     </div>`
     );
   });
--- a/src/formatter/formatTreeNode.js
+++ b/src/formatter/formatTreeNode.js
@@ -4,7 +4,7 @@ import formatReactElementNode from './formatReactElementNode';
 import type { Options } from './../options';
 import type { TreeNode } from './../tree';
 
-const jsxStopChars = ['<', '>', '{', '}'];
+const jsxStopChars = ['<', '>', '{', '}', '\n'];
 const shouldBeEscaped = (s: string) =>
   jsxStopChars.some(jsxStopChar => s.includes(jsxStopChar));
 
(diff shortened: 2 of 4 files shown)


shouze:
ping @Spy-Seth @vvo for a review & discussion.

We agree that as detailed on https://facebook.github.io/jsx/ (search for JSXTextCharacter) only `<`, `>`, `{`, `}` characters should require to protect a string, so having an option to explicitly protect multiline string nodes named `protectMultilineStringNodes` or `protectJSXText` could make sense.


This lib seems to have 2 use cases:
- Pretty debugging/diff tool.
- JSX generation (to then generate static websites for example).

So, as discussed a few hours ago with @Spy-Seth it maybe raises the debate to some other option that could trigger on many existing options, think it about some kind of `strictMode` or `safeMode` option.

Spy-Seth:
For now I'm not sure on how we should solve this problem.

The proposed solution by @shouze solve the issue. But will add lot's of ugly escaping in the output. People that use this library for outputting a component for debug will find this really annoying.

We could protect this behavior by an options to allow people to choose their output. We start to have lot's of options and each one we add make more difficult the maintenance and the on boarding for new user. 

I'm starting to think that we could embrace the two known usages of the library and create one dedicated render function for each case:
- to debug component: `renderReactElementToReaddableJsxString(element, options)`
- to re-execute the output: `renderReactElementToExecutableJsxString(element, options)`

Or any other APi we could find usable:
```js
renderReactElement(element).toReaddableJsxString(options);
renderReactElement(element).toExecutableJsxString(options);
```

We have to keep exposing some render options (the number of space/tabs, the use of tabs instead of space, the line max length) toallow the user to format the output. We could use the same logic than prettier and expose only a few number of basic customization options. 

I think in the long on long term this could ease the development for us and make our user more happier in each use case. What did you think @vvo?

vvo:
Reviving this, if I understood it well, the issue is that our current output cannot be then used to be re-executed at at later point in time?

What is the usecase behind getting a JSX string to then use it to be re-executed? I need to understand this before I can give an opinion..

Spy-Seth:
> Reviving this, if I understood it well, the issue is that our current output cannot be then used to be re-executed at at later point in time?

This is exactly the discussion. 

I no not know for the others, but we (@shouze and myself) are doing this to build a static site generator application.

vvo:
I would go for again another option. That or create a static method escapeJSXString() for it to be re-executed later on.

WDYT?

Spy-Seth:
Did you mean to parse again the result and to format it again?

vvo:
I meant:

```
reactElementToJSXString({
  safeForExecution: true
});
```

Or escapeJSXString() that would take a readable one and escape it for use in execution.

Spy-Seth:
I'm good for an options 👍 

I think I will not merge PR and introduce the same behavior in another PR behind an option toggle

PR #201, opened 2017-10-01
fix(flow): Not up to date React type usage

Spy-Seth:
I do not know why this doesn't break on CI when we upgrade flow.


Spy-Seth:
I guess this is because we use flow 52 in the master 😛 I need more sleep 🛌

PR #205, opened 2017-10-08
fix(formatting): trailling space

### LINKED ISSUE #135: Spaces between words and inline elements are lost
If I start with this JSX:

```js
<div>
  foo <strong>bar</strong> baz
</div>
```

That renders to React element that with children like this:

```js
[ 
  'foo ',
  { type: 'strong', props: { children: [Object] }, .. },
  ' baz'
]
```

If I then pass that element through this module, I come out with this JSX — replacing some whitespaces with `*` to clarify:

```js
<div>
  foo*
  <strong>
    bar
  </strong>
  *baz
</div>
```

The problem is that with the reformatting, those whitespace are actually lost. End-of-line and beginning-of-line whitespace is not meaningful, so the output HTML is actually this:

```html
<div>
  foo<strong>bar</strong>baz
</div>
```

If introducing all the line breaks is the preference here, I think one fix would be to add spaces as strings, e.g.

```js
<div>
  foo{' '}
  <strong>
    bar
  </strong>
  {' '}baz
</div>
```

Spy-Seth:
Closes #135

BREAKING CHANGE: Trailling are now preserved. In some rare case, `react-element-to-jsx-string` failed to respect the JSX specs for the trailing space. Event is the space were in the final output. There were silentrly ignored by JSX parser. This commit fix this bug by protecting the trailing space in the output.

If we take the JSX:
```jsx
<div>
  foo <strong>bar</strong> baz
</div>
```

Before it was converted to (the trailing space are replace by `*` for the readability):
```jsx
<div>
  foo*
  <strong>
    bar
  </strong>
  *baz
</div>
```

Now there are preserved:
```jsx
<div>
  foo{' '}
  <strong>
    bar
  </strong>
  {' '}baz
</div>
```

### DIFF
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -556,6 +556,22 @@ describe('reactElementToJSXString(ReactElement)', () => {
     );
   });
 
+  it('reactElementToJSXString(<div>\nfoo <span>bar</span> baz\n</div>)', () => {
+    expect(
+      reactElementToJSXString(
+        <div>
+          foo <span>bar</span> baz
+        </div>
+      )
+    ).toEqual(`<div>
+  foo{' '}
+  <span>
+    bar
+  </span>
+  {' '}baz
+</div>`);
+  });
+
   it('reactElementToJSXString(<div a={[1, 2, 3, 4]} />', () => {
     expect(reactElementToJSXString(<div a={[1, 2, 3, 4]} />)).toEqual(
       `<div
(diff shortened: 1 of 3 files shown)


src/index.spec.js
```diff
@@ -722,7 +738,9 @@ describe('reactElementToJSXString(ReactElement)', () => {
 
   it('reactElementToJSXString(<div> {false} </div>)', () => {
     expect(reactElementToJSXString(<div> {false} </div>)).toEqual(
-      '<div>\n    \n</div>'
+      `<div>
+  {'  '}
+</div>`
```
Spy-Seth: I'm not fan of the output that we have in that case. But the previous one were not pretty... so I guess we have to look at it in a dedicated PR.

Spy-Seth: I do not like this output. But the previous one were also not pretty... so I guess we have to look at it in a dedicated PR.

vvo: we can always provide an option to remove trailing whitespace no?

PR #223, opened 2017-10-28
Fix yarn -> npm

lahmatiy:
Yarn is not in dependencies and no reason for this. Script fails when run with no global yarn. With npm works just fine.


Spy-Seth:
Nice catch 👍

PR #224, opened 2017-10-28
Fix warning about access to PropTypes using React 15.5+ (fixes #213)

### LINKED ISSUE #213: "Warning: Accessing PropTypes via the main React package is deprecated" for React 15.5+
With commit https://github.com/algolia/react-element-to-jsx-string/commit/73d60526345e06f433929a7485abe1638e2da3da some import were changed to:

```js
import * as React from 'react';
```

In this case babel uses `_interopRequireWildcard` which is using `for..in` to make a copy of module exports. This cause to warning when React 15.5+ is using:

```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
```

Because with React 15.5, importing PropTypes from React is deprecated in favour of importing from the new 'prop-types' package.

lahmatiy:
Explicit import of React's types to avoid star import that leads to warning about access to PropTypes.

### DIFF
--- a/src/parser/parseReactElement.js
+++ b/src/parser/parseReactElement.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import * as React from 'react';
+import React, { type Element as ReactElement } from 'react';
 import type { Options } from './../options';
 import {
   createStringTreeNode,
@@ -9,7 +9,7 @@ import {
 } from './../tree';
 import type { TreeNode } from './../tree';
 
-const getReactElementDisplayName = (element: React.Element<*>): string =>
+const getReactElementDisplayName = (element: ReactElement<*>): string =>
   element.type.displayName ||
   element.type.name || // function name
   (typeof element.type === 'function' // function without a name, you should provide one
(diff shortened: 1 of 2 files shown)


src/parser/parseReactElement.js
```diff
@@ -1,15 +1,16 @@
...
 
-import * as React from 'react';
-import type { Options } from './../options';
+import React from 'react';
 import {
   createStringTreeNode,
   createNumberTreeNode,
   createReactElementTreeNode,
 } from './../tree';
+import type { Element as ReactElement } from 'react';
```
Mavrin: You could use one import from react
import React, { type Element as ReactElement } from 'react';

lahmatiy: Yep, that's work. Awesome! Thanks a lot!

Spy-Seth: `import React, { type Element as ReactElement } from 'react';`

Ho that's a good one! I didn't now it :D

Spy-Seth:
Good job 👍

PR #225, opened 2017-10-28
Remove `lodash` as unused dependency

lahmatiy:
Remove`lodash` from dependencies since is not in use


Spy-Seth:
I totally agree for the lodash removal 👍 

But I'm not sure about the removal of `collapse-white-space`. Could make a PR for each change? I need to double check why we need it in the first place 😃

lahmatiy:
Done

PR #231, opened 2017-11-07
Add commintlint

Spy-Seth:


vvo:
Wo seems awesome, thanks!

PR #242, opened 2017-11-10
feat(browser): build a dedicated version for the browser

Spy-Seth:
BREAKING CHANGE: This PR change of the internal directory structure of the exported code. The previous code has move from the `dist/` into the `dist/esm` directory (but remender that we do not avice you to do this 🤓


OriR:
@Spy-Seth Thanks for addressing this so quickly!

I see that the output bundle contains parts of react code.
This means that in case I upgrade my version of react that has changes in these code parts `react-element-to-jsx-string` won't use the newer version but rather the contained one.
This probably applies to all other dependencies, while the problem was only with a specific one.

Since I'm not familiar with rollup that much, is there a way to define an external dependency in rollup like there there is in [webpack](https://webpack.js.org/configuration/externals/)?
Or if there's something similar to [babel-engine-plugin](https://github.com/SamVerschueren/babel-engine-plugin) for rollup so that only the required libraries are actually inlined for the `cjs` build.

That way you could just define the packages you depend on and the package consumer would have to supply them, making sure there's less code duplication and is no longer version dependent but rather API dependent.

Spy-Seth:
You right, I've forgotten to mark react, and react-dom has an external dependencies. I will add it. 

I not sure of my choice of rollup. I find it more efficient for library bundling. But so few people know him... Maybe webpack could be a more reasonable tool.

Spy-Seth:
It's lighter without `react` and `react-dom` inside it 😜 
```
$ ls -lh dist/cjs/
total 56
-rw-r--r--  1 spy-seth  staff    22K Nov 13 09:57 bundle.js
-rw-r--r--  1 spy-seth  staff   823B Nov 13 09:57 bundle.js.map
```

vvo:
Looks good, any blog post or article explaining all the reasoning behind the various build endpoints in package.json?

Spy-Seth:
You could have a look onto the rollup wiki for the `module` entry: https://github.com/rollup/rollup/wiki/pkg.module

For the `browser` one, there is a reference on it in [rollup-plugin-node-resolve](https://github.com/rollup/rollup-plugin-node-resolve) readme 

By re-reading the `rollup-plugin-node-resolve` readme, I think I have made a mistake and we should target the esm build here and not the cjs one. Could someone give me a second thought on this?

```
      // some package.json files have a `browser` field which
      // specifies alternative files to load for people bundling
      // for the browser. If that's you, use this option, otherwise
      // pkg.browser will be ignored
      browser: true,  // Default: false
```

vvo:
maybe @Haroenv can help?

vvo:
On this message https://github.com/algolia/react-element-to-jsx-string/pull/242#issuecomment-343892518, both you and @samouss have rollup experience and might help @Spy-Seth here

Spy-Seth:
Are we confident on this change to directly do a major upgrade release. Or should we make some RC before?

Spy-Seth:
I just fix a typo in the first commit message

vvo:
Do a major, we don't have any community nowaday to ask to test that for an RC I guess. What do you think?

vvo:
You could always test it on your own project and the person having the browser need could also test it. If you release it under a different tag.

Spy-Seth:
I will try to find some time this weekend for some test and we could release it in the next week 😃

briandipalma:
Any chance this could be merged without adding tests?

Spy-Seth:
@vvo & @Haroenv All green and rebased :)

Sorry for the time to finish this

Haroenv:
Thanks for taking the time @Spy-Seth, Vincent is currently off, but I'll remind him to review this when he's back

danoc:
Hello! Would love to use this to make my site work in IE 11.

Anything I can do help out?

briandipalma:
I think we're just waiting for it to be published.

vvo:
@Spy-Seth Can I release the lib safely?

Spy-Seth:
Yes we could release it a major :)

PR #250, opened 2017-11-13
fix(formatting): Date/RegExp values output by formatComplexDataStructure

lahmatiy:
`sortobject` breaks output of Date and RegExp values, since makes an object copy of those values instead returning as is. Therefore, `formatComplexDataStructure` for following object

```js
{
    date: new Date('2017-11-13T00:00:00.000Z'),
    regexp: /test/g
}
```

outputs

```js
{
    date: {},
    regexp: {}
}
```

We can fix `sortobject` by sending a PR. But its [implementation is too verbose](https://github.com/bevry/sortobject/blob/master/source/index.js) for such simple task. So I decided to add an own implementation of `sortObject` instead of patching `sortobject`. This also helps keep things simpler and allows adjust behaviour if needed.

### DIFF
--- a/src/formatter/formatComplexDataStructure.spec.js
+++ b/src/formatter/formatComplexDataStructure.spec.js
@@ -86,4 +86,20 @@ describe('formatComplexDataStructure', () => {
   it('should format an empty array', () => {
     expect(formatComplexDataStructure([], false, 0, options)).toEqual('[]');
   });
+
+  it('should format an object that contains a date', () => {
+    const fixture = { a: new Date('2017-11-13T00:00:00.000Z') };
+
+    expect(formatComplexDataStructure(fixture, true, 0, options)).toEqual(
+      `{a: new Date('2017-11-13T00:00:00.000Z')}`
+    );
+  });
+
+  it('should format an object that contains a regexp', () => {
+    const fixture = { a: /test/g };
+
+    expect(formatComplexDataStructure(fixture, true, 0, options)).toEqual(
+      `{a: /test/g}`
+    );
+  });
 });
--- a/src/formatter/sortObject.spec.js
+++ b/src/formatter/sortObject.spec.js
@@ -1,0 +1,45 @@
+/* @flow */
+
+import sortObject from './sortObject';
+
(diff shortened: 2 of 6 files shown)


src/formatter/formatComplexDataStructure.spec.js
```diff
@@ -86,4 +86,20 @@ describe('formatComplexDataStructure', () => {
   it('should format an empty array', () => {
     expect(formatComplexDataStructure([], false, 0, options)).toEqual('[]');
   });
+
+  it('should format an object that contains a date', () => {
```
Spy-Seth: I guess theses tests are more related to `sortObject` beavior than the `formatComplexDataStructure` one. We should move then in the `sortObject.spec.js` file

src/formatter/sortObject.spec.js
```diff
@@ -0,0 +1,37 @@
+/* @flow */
+
+import sortObject from './sortObject';
+
+describe('formatComplexDataStructure', () => {
```
Spy-Seth: Oups! Should be `describe('sortObject', () => {`

src/formatter/sortObject.spec.js
```diff
@@ -0,0 +1,37 @@
...
+      c: date,
+    };
+
+    expect(sortObject(fixture)).toEqual({
+      a: [date, regexp],
+      b: regexp,
+      c: date,
+    });
+  });
+});
```
Spy-Seth: Could you add a test that validate that `sortObject` could be directly call with array `sortObject([{...}, {...}])`;

src/formatter/formatComplexDataStructure.spec.js
lahmatiy: `sortObject` has its own test for that. I think left it here since not only `sortObject` is used to process objects and other logic can break things too.

src/formatter/sortObject.spec.js
lahmatiy: Done

lahmatiy:
Done :)

PR #251, opened 2017-11-16
qa(formatting): Create a prettier config file to allow IDE to guess the used configurations

Spy-Seth:


vvo:
In the last eslint-config-algolia, prettier is bundled inside our eslint configuration. So when eslint passes, prettier passes too, in command line or even in IDEs.

The issue is then that if you have a prettier integration in your IDE, it will fail because the IDE has no way to know there's prettier bundled in eslint. So usually I only have the eslint plugin in IDEs, set with autofix.

If there are new ways to have both prettier and eslint that could be great. Let use know!

Spy-Seth:
I see a lot project separating prettier from eslint and using them separately. Not sure is better on a daily usage but it's a simpler installation.

For this PR I have only move the CLI arguments `"prettier:fix": "prettier --write --single-quote --trailing-comma es5 \"{src/**/*.js,package.json}\"` into a prettier config file. If these argument where fine there it should change nothing. But with the prettier/eslint integration maybe this script `prettier:fix` is not mandatory and a `"lint:fix": "eslint --fix ."` could be enough

PR #265, opened 2017-11-29
feature(format): support react fragment

Spy-Seth:
See https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

- [x] Upgrade docs
- [x] Add a body to the commit message to describe the change


KaboomFox:
this is awesome!

src/parser/parseReactElement.js
```diff
@@ -1,14 +1,17 @@
...
 import type { Options } from './../options';
 import {
   createStringTreeNode,
   createNumberTreeNode,
   createReactElementTreeNode,
+  createReactFragmentTreeNode,
 } from './../tree';
 import type { TreeNode } from './../tree';
 
+const supportFragment = Boolean(Fragment);
```
vvo: How does this works in previous versions, Fragment would be undefined or? I thought if the export is not present it would throw.

vvo:
Impressive PR, I can see the value of you having completely rewritten the library, it seems "easy" to add a new type of formatter. Congrats!

src/parser/parseReactElement.js
Spy-Seth: I was convinced that it return `undefined`, now you spoke about an error I have a doubt. 

After some test in node:
```js
import { Foo } from 'react';
console.log(Foo); // undefined
```

But I'm not sure this really friendly with the bundler (webpack, rollup...). They could have good reason to throw error. We could try like this and I will check with the PR #242 if this break or not. If it break we could try to add a package like `node-semver` to check the `React.version` value or by releasing a major to force people to use React `17.2` with the new version of lib. Will see 😄

vvo:
Would be good to test this with React 15 maybe

Spy-Seth:
I like that you said that. I've a PR that introduce some smoke test on multiple React version 😉  I will try make a move on them before finishing this PR

Spy-Seth:
I've rebase this PR. It seems that everything is working with react v15.6 and v16.1 😺

vvo:
👌 good to go

bdwain:
awesome!

bdwain:
any plans to merge this @Spy-Seth ?

KaboomFox:
What's the plan for this?

vvo:
@Spy-Seth How do you want to deal with the various PRs, do you need any particular help or advice? It seems this could just be shipped, let me know

Spy-Seth:
The only think that prevent me to finish these PR is that I'm temporary dev compatible computer-less. In a few weeks this should be resolved and I could works on this again 😊 sorry for the time and the lack of feedbacks

vvo:
> dev compatible computer-less.

Not sure to get that well BUT it's ok now I know the state, get back when you can, thanks

src/formatter/formatReactFragmentNode.js
```diff
@@ -0,0 +1,73 @@
...
+import type { Key } from 'react';
+import formatReactElementNode from './formatReactElementNode';
+import type { Options } from './../options';
+import type {
+  ReactElementTreeNode,
+  ReactFragmentTreeNode,
+  TreeNode,
+} from './../tree';
+
+const REACT_FRAGMENT_TAG_NAME_SHORT_SYNTHAX = '';
```
lahmatiy: `SYNTHAX` -> `SYNTAX`

Spy-Seth: Good catch

PR #269, opened 2017-12-14
qa(test): test the library agains more react and node versions

Spy-Seth:
Do not be scare by the number of line of code, it's mostly some `yarn.lock` file

This PR introduce a new tests suite called `smoke`. It tests that the library seems to works with multiple version React (v0.14, v15 and v16): it only validate some basic usage.

It also rework the `travis.yml` file to test our code with multiple node versions (the last v7, the last LTS/v8, and the last v9)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,26 @@
+dist: trusty
+sudo: false
+
 language: node_js
-node_js: stable
-branches:
-  only:
-    - master
-    - /^greenkeeper\/.*$/
+
+node_js: node # lastest
+
 cache: yarn
 
-# Since the upgrade to Ubuntu Precise, yarn is not available to run scripts: `yarn run foo`
-# So let's install it manually
 before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  # Install the latest yarn version
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s
   - export PATH="$HOME/.yarn/bin:$PATH"
+  - yarn --version
+
+install:
+  - yarn install
+
+script:
+  - yarn run test
+  - yarn run lint
+  - yarn run flow
+  - yarn run smoke 15.6.2
+  - yarn run smoke 16.1.0
+  - yarn run smoke default
+  - yarn run smoke next
--- a/package.json
+++ b/package.json
@@ -12,12 +12,16 @@
     "prepublish": "npm run build",
     "prebuild": "mkdir -p dist && rm -rf ./dist/*",
     "prettier:fix": "prettier --write \"**/*.{js,json}\"",
-    "test": "jest && npm run lint && npm run flow",
+    "test": "jest",
     "test:watch": "jest --watch",
-    "release": "./release.sh"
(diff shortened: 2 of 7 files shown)


tests/smoke/index.js
```diff
@@ -0,0 +1,11 @@
+const execFileSync = require('child_process').execFileSync;
+const path = require('path');
+
+const runTests = directoryName =>
+  execFileSync(path.join(__dirname, directoryName, 'index.js'), {
+    stdio: 'inherit',
+  });
+
+runTests('react-v14');
```
vvo: I would not test v14 (I would only test stable + last major version), but I would add a test maybe on the @next tag of React, that can be interesting.

tests/smoke/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
vvo: I wonder if it would have been doable to not have to copy paste the same project three times. Maybe by having a single test project, where between each jest call you just npm install another react version?

tests/smoke/react-v14/index.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+
+const React = require('react');
+const reactElementToJsxString = require('./../../../dist/').default;
```
vvo: It's ugly but at the same time you are testing the build! Which is good.

vvo: Do we need ".default"? I did not remember that, maybe because I was always using it from a babel transpiled frontend app.

Might be good to be able to require it without default in commonJS

.travis.yml
```diff
@@ -1,13 +1,30 @@
+dist: trusty
```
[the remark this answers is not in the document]
Spy-Seth: It tell travis to use their `trusty` container. It's the more recent and faster container they have to run the tests: https://docs.travis-ci.com/user/reference/overview/

.travis.yml
```diff
@@ -1,13 +1,30 @@
...
+
 language: node_js
-node_js: stable
-branches:
-  only:
-    - master
-    - /^greenkeeper\/.*$/
+
+node_js:
+  - "7"
```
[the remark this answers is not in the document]
Spy-Seth: You right for the node v7 version, I could replace it by the v6 one. 

This library mainly target node user, these test are to ensure our code does not break between node major release. But you right it's maybe too much since we do have that many code that use the node API directly. I will remove it and only test against the latest version

tests/smoke/react-v14/index.js
Spy-Seth: The `default` is mandatory. I have no idea if this possible to avoid using it. I will create a dedicated issue for this

Spy-Seth: See https://github.com/algolia/react-element-to-jsx-string/issues/271

.travis.yml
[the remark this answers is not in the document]
vvo: Ok I thought it was by default

vvo:
All good here?

Spy-Seth:
Yes!

Issue #285, opened 2018-04-27
Can not minify due to stringify-object dependency not targeting browsers


jonthomp:
I'm trying to use this within a [create-react-app](https://github.com/facebook/create-react-app/) project but am hitting build issues due to the dependency on [stringify-object](https://github.com/yeoman/stringify-object/).

The [issue is that they don't provide an ES5 compiled version](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify) and are [apparently not willing to do so](https://github.com/yeoman/stringify-object/issues/35#issuecomment-272376517).

Any chance of changing to an alternative dependency or just building in a compiled version of stringify-object rather than relying on the dependency?

jonthomp:
After trying out https://github.com/algolia/react-element-to-jsx-string/pull/242 it resolves the issue so I'll just be using that for now

danoc:
@jonthomp — Any luck? The latest version (14.0.0) doesn't seem to work with the older version of Uglify I'm using through Gatsby.

danoc:
I believe this is also related to: https://github.com/algolia/react-element-to-jsx-string/issues/211

vvo:
@DanielHoffmann Can you use a more recent version of Uglify with Gatsby?

pascalduez:
Fixed in #288 and #292

danoc:
Can confirm that this now works with older versions of Uglify.

PR #288, opened 2018-05-26
fix(browser): Add IE 11 support

### LINKED ISSUE #211: Support for IE11
This currently the latest version of `stringify-object` unfortunately it isn't published with an ES5 compiled version. 
I opened an [issue](https://github.com/yeoman/stringify-object/issues/53) there to publish an ES5 compliant version.
In case this won't get approved, I'm willing to make a PR to make the required changes to support IE11 😃

danoc:
This adds IE 11 support as well as support for older versions of Uglify.

It does so by transpiling `stringify-object` with babel.

Closes #211, #285

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -25,7 +25,9 @@ export default {
   plugins: [
     babel({
       babelrc: false,
-      exclude: 'node_modules/**', // only transpile our source code
+      // Don't transpile `node_modules` except for `stringify-object`. This enables IE 11 support
+      // and minification in older versions of Uglify.
+      exclude: 'node_modules/!(stringify-object)/**',
       presets: [
         [
           'es2015',


danoc:
The glob pattern doesn't seem to work. It's transpiling everything in `node_modules`.

Does anyone have tips on how to make it only exclude `stringify-object`?

pascalduez:
```js
exclude: 'node_modules/!(stringify-object)/**'
```

danoc:
Great, that worked well. Here's a diff of the CJS output:

https://www.diffchecker.com/RPnK1ARg

Notice that only `stringify-object` is affected.

danoc:
Thanks @Spy-Seth! @vvo — I think this one is good to go!

PR #290, opened 2018-05-28
fix(build): missing babel helpers, true esm modules, simplify

### LINKED ISSUE #289: Babel helpers missing
Hi,

recent changes in #242 introduced the use of the [`babel-plugin-external-helpers`](https://github.com/algolia/react-element-to-jsx-string/blob/master/rollup.config.js#L41). 
This yet produce exceptions `babelHelpers is not defined` since it's expecting it on the global object.

I use the lib bundled with webpack

EDIT: There seem to be several issues actually. The `esm` build is still compiled to commonjs.

pascalduez:
Hi,

- The Rollup Flow plugin is unnecessary, this is already handled by Babel and the `babel-preset-flow`.
- The current `esm` build is still transformed to `commonjs`, it's actually built two times.
So this simplify a bit things, and leverage Rollup for what it's good at (handling multiple builds).
- The source files with Flow annotations should be output into the `cjs` dir, flow-bin is a node tool.
- Don't exlude babel helpers from final bundle. Fixes #289 #294 
Only the `typeof` helper is used, so including it is amost costless. Including all helpers would,
or requiring the users to install the plugin and including it in their code.


pascalduez:
@vvo I think it's good to go.

pascalduez:
Smoke tests are failling, I could use some light here, what's the purpose of those? 

There's some weird things in:
```js
  if (buildType === 'esm') {
    return require(`./../../dist/esm`).default;
```
Importing an ES module in node with commonjs...

ArtieReus:
This fix works for me. The given example from #294

    reactElementToJSXString(<p style={{ color: 'red' }}>Test</p>)

goes through without errors.

.travis.yml
```diff
@@ -24,7 +24,7 @@ script:
   - yarn run smoke esm 15.6.2
   - yarn run smoke cjs 16.1.0
   - yarn run smoke esm 16.1.0
-  - yarn run smoke cjs default
-  - yarn run smoke esm default
+  - yarn run smoke cjs latest
+  - yarn run smoke esm latest
```
Spy-Seth: 👍

PR #309, opened 2018-09-13
fix(formatting) Fix JSX element in prop object recursive loop

samwalshnz:
Now supports:

```jsx
<Component
	props={{
		subcomponent: <AnotherComponent />,
	}}
/>
```

Fixes #307

Issue #307, opened 2018-09-13
Element with prop object containing an element causes maximum call stack error


samwalshnz:
```
import toString from 'react-element-to-jsx-string';

const el = <Element 
	prop={{
		component: <Component />,
	}}
/>

toString(el)
```

Running the above seems throw the following error: 

```
Uncaught RangeError: Maximum call stack size exceeded
    at Array.sort (:80/native)
    at sortObject (index.js?a730:225)
    at eval (index.js?a730:227)
    at Array.reduce (<anonymous>)
    at sortObject (index.js?a730:225)
    at eval (index.js?a730:227)
    at Array.reduce (<anonymous>)
    at sortObject (index.js?a730:225)
    at eval (index.js?a730:227)
    at Array.reduce (<anonymous>)
```

Here's a demo: https://stackblitz.com/edit/react-element-to-jsx-string-error?file=index.js

thomasbertet:
Hi there, just to emphaze a bit this issue, it happened to me too & its quite annoying. Would love to see this in the next release, happy to help test this is you need to.

samwalshnz:
@vvo I hadn't realised how the versioning worked for this repo, and so this PR didn't include an npm version bump. Are we able to retrospectively add one?

vvo:
@samwalshnz versioning is done at realease time, no need to bump in PRs. Am I missing something else here?

PR #312, opened 2018-10-04
test(ReactElementToJsxString): Add test when dealing with element inside prop object

thomasbertet:
Following issue https://github.com/algolia/react-element-to-jsx-string/issues/307 & PR https://github.com/algolia/react-element-to-jsx-string/pull/309

I added two tests: 
- One is to cover the one-deep level of nesting, ie. an element inside a prop object.
- One is to over deep-nested level, .ie. an element inside a prop object inside another element inside a prop object ... you get the idea :dancer: 

Please tell me if anything is unclear or needs to follow certain rules. I tried to make them just as the others but I might have missed some. 

@vvo Thanks for your time! Would love to see this fix released


PatrickSachs:
Hello @thomasbertet,

thank you very much for your contribution.

Unfortunately this test does not cover the issue in #307. As can be seen in [this](https://stackblitz.com/edit/react-element-to-jsx-string-error?file=index.js) code snippet, the error only occurs if the React element to be transformed into a string was created by another React component.

This is because React adds the "_owner" key to these elements pointing to the owning React component, which can cause an infinite loop. 
If the React component was created outside of another components lifecycle, this key does not exist, and thus the stack overflow will not occur.

The unit tests are ran by Jest which is separate from the React lifecycle, thus elements created in the unit tests do not have the "_owner" key. This can be verified by (temporarily) commenting out the fix of #307 and then running the tests again.

I'm not sure if we can properly test this using our current testing setup without using a virtual dom.

thomasbertet:
Oh I see... I thought it was quite easier. Sorry about the misunderstanding. I could have done the check you did by commenting out the fix. 
Let me try some and I will get back to you.

PatrickSachs:
Without having looked at the tests too much yet, I'm not a fan of adding this .spec.js file in a different directory than the other ones. I think the test either belongs in `index.spec.js` or `sortObject.spec.js`.

The tests fail and pass as expected, but I'll have a more in depth look at it.

thomasbertet:
@PatrickSachs moved the test file.

PatrickSachs:
This looks good to me. Adding Enzyme for currently (just) a single unit tests is a big heavy, but I believe it will become useful for more tests in time.

Unless any more issues crop up related to this I'll create a new release tomorrow. @samwalshnz

PR #313, opened 2018-10-05
Attempt to Fix #299

### LINKED ISSUE #299: Dependency deprecation warning: babel-preset-es2015 (npm)
On registry `https://registry.npmjs.org`, the "latest" version (v6.24.1) of dependency `babel-preset-es2015` has the following deprecation notice:

`🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! `

Marking the latest version of an npm package as deprecated results in the entire package being considered deprecated, so contact the package author you think this is a mistake.

Affected package file(s): `package.json`

If you don't care about this, you can close this issue and not be warned about `babel-preset-es2015`'s deprecation again. If you would like to completely disable all future deprecation warnings then add the following to your config:

```
"suppressNotifications": ["deprecationWarningIssues"]
```

frytyler:
Hi, 

I saw #299 and tried to get it working however it turned out to be a fairly deep rabbit role. I seem to be stuck now anyone have any advice?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "stage-2", "react", "flow"]
+  "presets": ["@babel/preset-env", "@babel/preset-react", "@babel/preset-flow"]
 }
--- a/package.json
+++ b/package.json
@@ -35,16 +35,17 @@
   "license": "MIT",
   "repository": "algolia/react-element-to-jsx-string",
   "devDependencies": {
+    "@babel/cli": "^7.1.2",
+    "@babel/core": "7.1.2",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "@babel/preset-stage-2": "^7.0.0",
+    "@babel/register": "^7.0.0",
     "@commitlint/cli": "7.1.2",
     "@commitlint/config-angular": "7.1.2",
-    "babel-cli": "6.26.0",
     "babel-eslint": "10.0.1",
-    "babel-jest": "22.2.2",
-    "babel-preset-es2015": "6.24.1",
-    "babel-preset-flow": "6.23.0",
-    "babel-preset-react": "6.24.1",
-    "babel-preset-stage-2": "6.24.1",
-    "babel-register": "6.26.0",
+    "babel-jest": "^23.6.0",
     "conventional-changelog-cli": "2.0.5",
     "doctoc": "1.3.1",
     "eslint": "5.6.1",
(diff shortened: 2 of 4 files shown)


PatrickSachs:
Hello @frytyler!
Thank you very much for tackling this!
  
This is some really strange ES6/CJS module interop going on here.
The build currently fails with this message: `Error: 'default' is not exported by node_modules\stringify-object\index.js`
  
If one looks at the source code of this module, it turns out if does indeed not have a ES6 style default export, but instead exports its function directly.

```js
module.exports = (val, opts, pad) => {
	const seen = [];
// ...
```

This is all fine in CJS land, but we are running into a problem since `import stringify from 'stringify-object'` essentially translates to `const stringify = require('stringify-object').default`.

Normally this sort of module system interop should be handled by the bundler for us. (And it is correctly handled e.g. for the `is-plain-object` dependency, which is also using the same module system.)

Maybe @danoc can help out with this, since there is some special handling for `stringify-object` in [rollup.config.js#L31](https://github.com/algolia/react-element-to-jsx-string/blob/master/rollup.config.js#L31) by him?

PatrickSachs:
I personally see no issue with transitioning to Babel 7 as long as we don't cause any problems for the end users in the process.
However, I am not sure if I am authorized to make a decision on that. An actual member of the Algolia organization like @vvo might be able to give some insight on who gets to make these decisions.

Haroenv:
Since this is used as a dev dependency for users, I don't see why transitioning to babel 7 here would have an impact on users. It could be a major or breaking change if you want to be extra careful?

vvo:
As long as we compile down to the same platforms than today there's nothing wrong in upgrading all dependencies. Find out which platforms we supported before (most probably any platform supporting es2015) and compile down to the same platforms and you are good to go.

If later on you want to compile to only more modern platforms, that's also a possibility but the easiest is usually to do it in different steps and only when needed (for example for compile size issues or compatibility issues we might not have here).

vvo:
@PatrickSachs I trust your judgment here it's the right mindset (think about users first)

vvo:
thanks @Haroenv for jumping here too

Spy-Seth:
I'm trying to make the upgrade on rollup (see https://github.com/algolia/react-element-to-jsx-string/pull/341) and I stuck with the issue as you in this PR. I will try to find a solution on this

Spy-Seth:
Some new digging into the issue make found a workaround. It will be shipped by the PR https://github.com/algolia/react-element-to-jsx-string/pull/351

frytyler:
Can this be closed? I don't have time to continue the work here

Spy-Seth:
Thanks for your try!

PR #315, opened 2018-10-19
Fixed Flow typing to allow for omitting options when using the options param

FermiDirak:
Currently options are an all or nothing parameter. This PR splits the `Options` type to `OptionParams` and `Options` so that options can be independently specified when using the options param.


FermiDirak:
I'm open to adding tests

vvo:
@FermiDirak why not just updating the already in place Options type definition (warning: I am new to Flow)

FermiDirak:
@vvo Flow isn't able to infer default values in a function signature, so when you consume `reactElementToJsxString` and use Flow, the `Options` argument must be specified with all its properties defined. The problem with updating the `Options` type is that some fo the methods methods in the library require all of `options`s properties to be defined. The way to circumvent Flow not inferring defaults is to create a separate type for the function signature.

Spy-Seth:
To avoid the duplication you could use the [`$Shape` special keyword](https://flow.org/en/docs/types/utilities/#toc-shape)

Here an example:
https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDyOGAlnAHYDOYAvGAN4A+UccAXGBQK4C2ARngBOAGjD8AhkM782MPBIpMAvplwFiZSgAUpE3jXoASAMoALCfgA8m8tQB8qMGHQAKBqw5gALACZl7LY6egYAlADcbgyS0mAA5BRwGGAS4nIKFHEBQRS6QvpUEUA

Issue #286, opened 2018-05-10
API request: filterFunction


adamduncan:
The `filterProps` option is helpful, but lacks the ability to dynamically filter props based on anything other than their name.

What if the API included a `filterFunction` option, that would take propNames and allow us to pass our own filter function body? That way we could also filter out props based on more dynamic criteria. E.g. the prop's value, or the props type.

For example:

```
reactElementToJSXString(element, {
  filterProps: ['onClick'], // useful in scenarios when you know the specific prop names
  filterFunction: propName => typeof element.props[propName] !== 'function' // don't render props which are functions
  // or
  filterFunction: propName => !element.props[propName] // don't render props that have falsy values (e.g. empty strings)
})
```

Thanks for your work on this library!

vvo:
Good idea. I would go for `filter`, as a function yes, deprecate filterProps with a warning and in docs.

Accepting PRs to solve that @adamduncan gogogo!

PR #333, opened 2018-11-30
feat: allow filterProps to be a function

bschlenk:
filterProps can now be passed as either an array of strings, or a function that will be called with the props value and key and return false to filter out that prop.

Closes https://github.com/algolia/react-element-to-jsx-string/issues/286

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,13 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   Just return the name you want for the provided ReactElement, as a string.
 
-**options.filterProps: array, default []**
+**options.filterProps: array | (val: any, key: string) => boolean, default []**
 
-  Provide an array of props to filter for every component. For example ['key'] will suppress the key="" prop from being added.
+  If an array of strings is passed, filter out any prop who's name is in
+  the array. For example ['key'] will suppress the key="" prop from being added.
+
+  If a function is passed, it will be called with the prop value and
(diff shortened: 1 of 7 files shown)


vvo:
@bschlenk any idea why tests are not passing?

bschlenk:
I'm not sure, the tests pass on my local machine. The test that fails says "The propNameSorter should always move the `key` and `ref` keys first", but specifically asserts that `key` comes before `ref`. Is that the expected behavior? Something is causing the sort to be unstable.

bschlenk:
Now it is failing for some flow checks. I've never used flow, I'm trying to fix it but I keep getting 

```
Error ------ src/formatter/createPropFilter.js:10:19

Cannot call filter because array type [1] is not a function.

 [1]  5│   filter: string[] | ((any, string) => boolean)
      6│ ) {
      7│   if (Array.isArray(filter)) {
      8│     return key => filter.indexOf(key) === -1;
      9│   } else {
     10│     return key => filter(props[key], key);
     11│   }
     12│ }
     13│
```

I thought flow was supposed to be able to do type refinement based on the Array.isArray check?

Spy-Seth:
Now that we have fix most of our CI/build issues, we could work on this feature again :)

bschlenk:
Want me to rebase?

bschlenk:
I forgot I had a PR for this package and deleted my fork 😛
Closing this in favor of https://github.com/algolia/react-element-to-jsx-string/pull/417

PR #337, opened 2019-01-04
fix: testing with jest not working

adammockor:
fixes SecurityError: localStorage is not available for opaque origins


Spy-Seth:
Can you explain in witch case the `testURL` configuration is needed?

adammockor:
If you clean install this repo with `npm`, you will get `jsdom@11.12.0` from `jest` which have this issue https://github.com/jsdom/jsdom/issues/2304. Setting `testURL` will fix that, because jest by default has URL `about:blank` where local storage is not available.

Spy-Seth:
I could not reproduce the issue on a fresh install. Did you still have it? We do a lot of dependencies update during the last days

Spy-Seth:
I will close this PR as we have no reproduction. If someone have a the same I will be glad to open it again and have second thought on it.

PR #338, opened 2019-01-04
feat(formatting): render children as function

adammockor:


adammockor:
I will fix the flow errors. But failing test is not related to this PR and local `yarn run test` works without error.

vvo:
Yep I don't get why it's failing on travis only
image

vvo:
cc @Spy-Seth maybe?

DavidBabel:
Do you use the same node version as your travis build ?  `node_js: node # lastest`

src/parser/parseReactElement.js
```diff
@@ -71,6 +72,14 @@ const parseReactElement = (
...
     .map(child => parseReactElement(child, options));
 
+  if (typeof element.props.children === 'function') {
+    const functionChildrens = parseReactElement(
+      element.props.children(),
+      options
+    );
+    childrens.push(createReactFunctionTreeNode([functionChildrens]));
+  }
+
```
Spy-Seth: One gotcha with this detection method is is case of mixed children (or multiple function as a children):

```js
// Multiple functions as child
reactElementToJSXString(
  <div>
    {() => <div>Hello Foo</div>}
    {() => <div>Hello Bar</div>}
  </div>,
  {
    showFunctions: true,
  }
)

// Mixed content
reactElementToJSXString(
  <div>
    {() => <div>Hello Foo</div>}
    <span>Some other children</span>
  </div>,
  {
    showFunctions: true,
  }
)
```

I know this is should not be frequent or maybe event imaginable but the JSX allow it 🤷‍♂️

I do some try with you PR, it's a shame that `React.Children` filters out the function children. We do not want to re-implement it to be able to conserve function in our parsing.

adammockor: [This](https://github.com/facebook/react/issues/9821) and [this](https://github.com/facebook/react/issues/11888) is related  to `React.Children` behavior.

PR #340, opened 2019-01-05
fix(formatting): Make the props "key" and "ref" order predictibale

Spy-Seth:
...and fix the CI ^^'

### DIFF
--- a/src/formatter/propNameSorter.js
+++ b/src/formatter/propNameSorter.js
@@ -1,13 +1,17 @@
 /* @flow */
 
+const isKeyOrRefProps = (propName: string) => ['key', 'ref'].includes(propName);
+
 export default (sortProps: boolean) => (a: string, b: string): -1 | 0 | 1 => {
   if (a === b) {
     return 0;
   }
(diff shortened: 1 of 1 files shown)


vvo:
This is an interesting fix, can you teach me a bit more and explain why it fixes the issue?

Spy-Seth:
The order function was not stable. It return a different result depending on the order of the original props because we forgot to enforce the order between the `key` and the `ref` values:

Before the fix the output was not the same:
```js
console.log(['key', 'ref']).sort(propNameSorter)); // ['key', 'ref'] 
console.log(['ref', 'key']).sort(propNameSorter));  // ['ref', 'key']
```

After the fix the output is the same:
```js
console.log(['key', 'ref']).sort(propNameSorter)); // ['key', 'ref']
console.log(['ref', 'key']).sort(propNameSorter));  // ['key', 'ref']
```

Why this do not trigger before? I'm not sure: maybe some react internal change ?

vvo:
Thanks!

PR #351, opened 2019-01-27
chore(build): replace babel-preset-es2015 by @babel/preset-env

### LINKED ISSUE #299: Dependency deprecation warning: babel-preset-es2015 (npm)

Spy-Seth:
This also unlock the upgrade to jest v24 and rollup v1

This PR will fix #299, fix #304, fix #313, fix #334, fix #341 and fix #350

We will get rid of these staging PR 💪


src/formatter/formatComplexDataStructure.js
```diff
@@ -1,14 +1,17 @@
...
+import * as stringifyObject from 'stringify-object';
 import sortObject from './sortObject';
 import parseReactElement from './../parser/parseReactElement';
 import formatTreeNode from './formatTreeNode';
 import formatFunction from './formatFunction';
 import spacer from './spacer';
 import type { Options } from './../options';
 
+// Rollup fail to correctly import `stringify-object` so we have to use a workaround
+const stringify = (...args) => stringifyObject.default(...args);
```
Spy-Seth: The `import * as stringifyObject from 'stringify-object';` and this workaround allow us to avoid the rollup issue where @frytyler get stuck into in PR https://github.com/algolia/react-element-to-jsx-string/pull/313

Spy-Seth:
- [ ] Compare the bundle before/after this PR to ensure everything is OK

Issue #344, opened 2019-01-14
React 16 createRef crash


neutraali:
Turns out if a component has any refs created with **React 16+** the lib crashes. For example, trying to `react-element-to-jsx-string` the following component:

```javascript
class Example extends React.Component {
	constructor(props) {
		super(props);
		this._ref = React.createRef();
	}
	render() {
		return <div ref={this._ref} />;
	}
}
```
... Will result in the following crash:
```
Uncaught RangeError: Maximum call stack size exceeded
    at Array.reduce (<anonymous>)
    at sortObject (index.js:225)
    at eval (index.js:230)
    ...
```
A (hopefully) temporary fix is to filter out said props, but I don't think this is ideal.
```javascript
this.reactElementToJSXString(item, {
	filterProps: ['ref', 'key']
	...
});
```

Spy-Seth:
Hi @neutraali 

Thanks for reporting this issue. 

I have try reproduce it in our test suite but without success:

```js
it('reproduce the issue #344', () => {
  const Example = class extends React.Component {
    constructor(props) {
      super(props);
      this._ref = React.createRef();
    }
    render() {
      return <div ref={this._ref} />;
    }
  }

  console.log(reactElementToJSXString(<Example />)); // Output: "<Example />"
})
```

Could you give us more detail ? Exact React version, environment, or more code to reproduce it

neutraali:
Thanks for the speedy reply @Spy-Seth. Here's some package information:
```
"react": "16.4.1"
"react-dom": "16.4.1"
"react-element-to-jsx-string": "14.0.2"
```
We're using the following options for `reactElementToJSXString` (excluding `filterProps`, which we're using to circumvent the bug):
```
showDefaultProps: false,
showFunctions: true,
sortProps: false,
maxInlineAttributesLineLength: 100,
tabStop: 4
```
In our specific situation, we have a component that calls `reactElementToJSXString` to all it's children, so I'm assuming the bug doesn't manifest unless one explicitly passes the prop to the function:

```javascript
const RendererExample = class extends React.Component {
	render() {
		return <div>
			{reactElementToJSXString(this.props.children)}
		</div>;
	}
}

const WrapperExample = class extends React.Component {
	constructor(props) {
		super(props);
		this._ref = React.createRef();
	}
	render() {
		return <RendererExample>
			<Example ref={this._ref} />
		</RendererExample>
	}
}
```
[...]

PR #357, opened 2019-03-10
fix: Prevent infinite loop with React 16 refs

skvale:
Relates to #344 	

This provides a reproduction of the bug with a test.

Remove this block
```jsx
 if (key === 'current') {
        // eslint-disable-next-line no-param-reassign
        result[key] = '[Circular]';
 }
```
and it will fail with the `Maximum call stack size exceeded` going into the circular 
```
element->ref->current->element->ref->current
```
loop.


vvo:
Hey there, thanks, I was trying to fix your tests, and thought that it was because of incompat issues with latest nodejs, so I created: https://github.com/algolia/react-element-to-jsx-string/pull/358

but then there's another test error we have seen in the past already (https://github.com/algolia/react-element-to-jsx-string/pull/340) cc @Spy-Seth à l'aide :D (help us)

Spy-Seth:
😭😭😭😭

I will not able to have a look this week 😢

Spy-Seth:
Fixing #358 and rebasing this PR make it green. Not this a good thing: it means our code try to write on read only object attributes sometimes and node v12 now trigger error on CI (not on my laptop...). We should investigate this

hipstersmoothie:
any progress on this?

hipstersmoothie:
Actually i just fixed it for the mean time. All you have to do is add `.toString = () => '{} => {}';` to the `current` of your ref

vvo:
I see this code is now working on Node v12 so we're good to go

vvo:
released as 14.0.3

Issue #387, opened 2019-07-23
CJS output isn't ES5


adierkens:
The `/dist/cjs/index.js` file contains arrow functions which aren't commonjs

```
var lib = createCommonjsModule(function (module, exports) {
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = (object) => Object
    .getOwnPropertySymbols(object)
    .filter((keySymbol) => object.propertyIsEnumerable(keySymbol));

});
```

This causes the project to break support on older browsers (like IE)

It looks like in `14.0.2` this worked: 
```
var lib = createCommonjsModule(function (module, exports) {
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = function (object) { return Object
    .getOwnPropertySymbols(object)
    .filter(function (keySymbol) { return object.propertyIsEnumerable(keySymbol); }); };

});
```

ahuth:
Seeing the same. Seems that this project's dependency `stringify-object` updated its own dependency on `get-own-enumerable-property-symbols` which packages an arrow function.

Perhaps this package can downgrade `stringify-object`?

Spy-Seth:
We could yes. `stringify-object` is a pain point for us if it could be deleted it could be perfect

hipstersmoothie:
Yeah we saw this too and is causing use to tweak out webpack configs to include this library

Spy-Seth:
https://github.com/algolia/react-element-to-jsx-string/pull/408 is merged and release in [v4.1.0](https://github.com/algolia/react-element-to-jsx-string/releases/tag/v14.1.0) 🎉

PR #408, opened 2019-09-02
Fix output for es5

Chris-Baker:
Fix for #387 to replace stringify-object and make the output compatible with ES5.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,8 +83,7 @@
   },
   "dependencies": {
     "@base2/pretty-print-object": "^1.0.0",
-    "is-plain-object": "3.0.0",
-    "stringify-object": "3.3.0"
+    "is-plain-object": "3.0.0"
   },
   "jest": {
     "setupTestFrameworkScriptFile": "<rootDir>tests/setupTests.js"
--- a/package.json
+++ b/package.json
@@ -23,10 +23,7 @@
     "smoke": "node tests/smoke/run"
   },
   "lint-staged": {
-    "*.js": [
-      "prettier --write \"**/*.{js,json}\"",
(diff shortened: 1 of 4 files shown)


vvo:
@Spy-Seth this seems good for me, tests are passing 🤷‍♂ thoughts?

Spy-Seth:
I'm not fan to use a community fork of `stringify-object` only dedicated to this package. If we use a fork we should manage it (as there is no string alternative package). @vvo Could we fork it and publish it from the Algolia org?

Spy-Seth:
To clarify my answer from yesterday, I'm all in favor of resolving the `stringify-object` bundling issue. It's a pain in the ass for us for too long: source of bugs in the browser and maintenance headaches (we are in the rollup upgrade because of it...)

Chris-Baker:
Hi @Spy-Seth, @vvo , let me know if you want me to make changes to this once you have decided whether to publish an internal fork.

vvo:
> @vvo Could we fork it and publish it from the Algolia org?
I would rather use the one from @Chris-Baker since it seems he had the issue and he would be the best to maintain it. There's also https://github.com/searls/stringify-object/tree/transform-es5 which already existed too if we want to reuse something and if @Chris-Baker want to work with @searls.

Latest release from stringify-object was one year ago, I believe it's fine to just merge this if the only issue is about who owns the fork. Or reuse https://github.com/searls/stringify-object/tree/transform-es5.

@Spy-Seth I won't be at Algolia forever and don't want Algolia to own something they did not need (stringify object es5) but rather trust the community (@Chris-Baker) to do good.

PR #417, opened 2019-09-18
feat: allow filterProps to be a function

### LINKED ISSUE #286: API request: filterFunction

bschlenk:
filterProps can now be passed as either an array of strings, or a function that will be called with the props value and key and return false to filter out that prop.

Closes #286

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,13 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   Just return the name you want for the provided ReactElement, as a string.
 
-**options.filterProps: array, default []**
+**options.filterProps: string[] | (val: any, key: string) => boolean, default []**
 
-  Provide an array of props to filter for every component. For example ['key'] will suppress the key="" prop from being added.
+  If an array of strings is passed, filter out any prop who's name is in
+  the array. For example ['key'] will suppress the key="" prop from being added.
+
+  If a function is passed, it will be called for each prop with two arguments,
+  the prop value and key, and will filter out any that return false.
 
 **options.showDefaultProps: boolean, default true**
 
--- a/src/formatter/createPropFilter.js
+++ b/src/formatter/createPropFilter.js
@@ -1,0 +1,12 @@
+/* @flow */
+
+export default function createPropFilter(
+  props: {},
+  filter: string[] | ((any, string) => boolean)
+) {
+  if (Array.isArray(filter)) {
+    return (key: string) => filter.indexOf(key) === -1;
+  } else {
+    return (key: string) => filter(props[key], key);
+  }
+}
--- a/src/formatter/formatReactElementNode.js
+++ b/src/formatter/formatReactElementNode.js
@@ -5,6 +5,7 @@ import formatTreeNode from './formatTreeNode';
 import formatProp from './formatProp';
 import mergeSiblingPlainStringChildrenReducer from './mergeSiblingPlainStringChildrenReducer';
 import sortPropsByNames from './sortPropsByNames';
+import createPropFilter from './createPropFilter';
 import type { Options } from './../options';
 import type { ReactElementTreeNode } from './../tree';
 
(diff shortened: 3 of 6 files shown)


Spy-Seth:
Thanks for rebasing this PR 👍 I will have a look into it in the next days

src/formatter/formatReactElementNode.js
```diff
@@ -126,13 +127,15 @@ export default (
 
   const visibleAttributeNames = [];
 
+  const propFilter = createPropFilter(props, filterProps);
```
Spy-Seth: The `propFilter` variable could be renamed `filterOutBlacklistedProps` to be more explicit on what it does

README.md
```diff
@@ -70,9 +70,13 @@ console.log(reactElementToJSXString(<div a="1" b="2">Hello, world!</div>));
 
   Just return the name you want for the provided ReactElement, as a string.
 
-**options.filterProps: array, default []**
+**options.filterProps: string[] | (val: any, key: string) => boolean, default []**
```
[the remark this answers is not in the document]
bschlenk: Is there a good enough reason to deprecate the string array form? It would be trivial for users to switch to the function, but the string form makes what I believe to be the common case pretty simple.

src/formatter/createPropFilter.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+
+export default function createPropFilter(
+  props: {},
+  filter: string[] | ((any, string) => boolean)
+) {
+  if (Array.isArray(filter)) {
+    return (key: string) => filter.indexOf(key) === -1;
+  } else {
+    return (key: string) => filter(props[key], key);
```
bschlenk: Do we agree on the filter function being called with the prop value as the first argument, and the prop key as the second? I did it that way only because that's how most of lodash's functions work ([lodash#filter](https://lodash.com/docs/4.17.15#filter), [lodash#map](https://lodash.com/docs/4.17.15#map)), and I'm used to lodash.

If the string array form is really going to be deprecated, then it might make more sense for the key to be the first argument here. Then you wouldn't have to write your filter function with an unused first arg if you just wanted to filter by key.

src/formatter/formatReactElementNode.js
bschlenk: I can change it, although I don't think it's any more explicit really. With the function form there isn't a list of any sort,  and even then it's more of a "whitelist" since you return `true` to keep the value.

leonelgalan:
Let's get this merged! I was looking for a way to filter props with value of `undefined`, having a function allows for this and more.

Spy-Seth:
Let's do this! 🚀

Spy-Seth:
Released in v14.3.0

elliotgeno:
I see this is still relatively fresh!!
I would like to make one improvement...  Instead of optionally showing the prop with a function, I would like to override the string for the prop's value with that the filterProps function.
For example, if you provide a prop with a reference to another Component, it will try to render that as an object.
Example:
`<Accordion.Toggle
      as={Card.Header}
    >`
returns this string: 
`<Accordion.Toggle
      as={{
        $$typeof: Symbol(react.forward_ref),
        defaultProps: undefined,
        displayName: 'Card.Header',
        render: function noRefCheck() {}
      }}
    >`

Ideally, I could write a function that replaces that prop value with the component's displayName. Similar to how options.displayName works... but for property values.

Issue #414, opened 2019-09-15
Test files are published in each release


Spy-Seth:
```
npm notice
npm notice 📦  react-element-to-jsx-string@14.1.0
npm notice === Tarball Contents ===
npm notice 2.7kB  package.json
npm notice 34B    .prettierignore
npm notice 411B   .prettierrc
npm notice 20.4kB CHANGELOG.md
npm notice 65B    commitlint.config.js
npm notice 1.1kB  LICENSE
npm notice 6.2kB  README.md
npm notice 46B    renovate.json
npm notice 1.1kB  rollup.config.js
npm notice 1.2kB  .vscode/launch.json
npm notice 69B    .vscode/settings.json
npm notice 212B   dist/cjs/AnonymousStatelessComponent.js.flow
npm notice 1.5kB  dist/cjs/formatter/formatComplexDataStructure.js.flow
npm notice 3.4kB  dist/cjs/formatter/formatComplexDataStructure.spec.js.flow
npm notice 607B   dist/cjs/formatter/formatFunction.js.flow
npm notice 1.6kB  dist/cjs/formatter/formatFunction.spec.js.flow
npm notice 1.6kB  dist/cjs/formatter/formatProp.js.flow
npm notice 5.2kB  dist/cjs/formatter/formatProp.spec.js.flow
npm notice 1.7kB  dist/cjs/formatter/formatPropValue.js.flow
npm notice 4.0kB  dist/cjs/formatter/formatPropValue.spec.js.flow
npm notice 5.1kB  dist/cjs/formatter/formatReactElementNode.js.flow
npm notice 2.6kB  dist/cjs/formatter/formatReactElementNode.spec.js.flow
npm notice 1.7kB  dist/cjs/formatter/formatReactFragmentNode.js.flow
npm notice 2.7kB  dist/cjs/formatter/formatReactFragmentNode.spec.js.flow
npm notice 253B   dist/cjs/formatter/formatTree.js.flow
npm notice 487B   dist/cjs/formatter/formatTree.spec.js.flow
npm notice 1.3kB  dist/cjs/formatter/formatTreeNode.js.flow
npm notice 1.6kB  dist/cjs/formatter/formatTreeNode.spec.js.flow
npm notice 795B   dist/cjs/formatter/mergeSiblingPlainStringChildrenReducer.js.flow
npm notice 2.4kB  dist/cjs/formatter/mergeSiblingPlainStringChildrenReducer.spec.js.flow
[...]

PR #437, opened 2019-10-06
chore: ignore test files when publishing

tisoap:
Looking for simple issues to help with for Hacktoberfest, I stumbled upon this repo :slightly_smiling_face:

This PR fixes issue #414

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,5 +5,7 @@
 .travis.yml
 src/
 tests/
+**/*.spec.js
+**/*.spec.js.flow
 !dist/
 release.sh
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -4,5 +4,6 @@
 .flowconfig
 .travis.yml
 src/
+tests/
```
Spy-Seth: Can you also add the `**/*.spec.js` and `**/*.spec.js.flow` patterns to also prevent the unit tests files to be published?

tisoap: Added!

PR #475, opened 2019-11-22
feat(typescript): Add TypeScript declaration file. Closes #474

### LINKED ISSUE #474: Add a TypeScript declaration file
Hey there 👋 

This module is really helpful, so first thanks for creating and maintaining it!

Unfortunately, there is no TypeScript declaration file included, and AFAICS there is also no typings package on `@types`. Could you imagine adding such a file to this module?

This would help a lot when this module is used from TypeScript.

PS: I could also imagine to come up with a PR, but I have seen that you are using flow, and I was not sure if you would like to have a TypeScript declaration file at all, so I just wanted to ask first.

goloroden:
Hey there 👋 

As suggested in #474 there is currently a TypeScript declaration file missing. Hence I added one. If you want to, please feel free to merge it 😊


goloroden:
Is there any chance that this will get merged eventually?

PR #543, opened 2020-03-22
fix: forward ref not working

fat:
rebased… and added with yarn…


fat:
This rebases + fixes the commit message for https://github.com/algolia/react-element-to-jsx-string/pull/461

fat:
sad

Andarist:
@armandabric I'm pretty sure that this can be closed now since #617 has landed

PR #579, opened 2020-10-19
correctly acquire name of forward ref's react element fix: forwardRef

dvnrsn:


src/parser/parseReactElement.js
```diff
@@ -13,6 +13,7 @@ import type { TreeNode } from './../tree';
 const supportFragment = Boolean(Fragment);
 
 const getReactElementDisplayName = (element: ReactElement<*>): string =>
+  (element.type.render && element.type.render.name) ||
```
dvnrsn, line 16: image
In the case of a type `react.forward_ref`, the render has the name
image

Andarist:
@armandabric I'm pretty sure that this can be closed now since #617 has landed

PR #616, opened 2021-09-21
fix: handle circular references in `sortObject`

aleclarson:
Instead of handling circular references for React refs only, handle them for everything.

Published under [`@alloc/react-element-to-string`](https://www.npmjs.com/package/@alloc/react-element-to-string) for anyone who needs this change now.


aleclarson:
I switched to [`pretty-format`](https://www.npmjs.com/package/pretty-format), which has `maxDepth` control and plugin support, so I won't be following through with this PR.

Issue #352, opened 2019-02-12
DisplayName not retreived by default from latest React APIs


pascalduez:
```js
function Foo() {
  return <div />
}

export default React.memo(Foo);
```

Results in:
```
<[object Object] />
```

See https://github.com/facebook/react/issues/14319

Opening an issue here to track the progress, and see what we can do.  
The thing is we need the displayName not only for DEV env, but also production.

Spy-Seth:
We are always happy to receive PR 😄

mallow-fight:
when to fix this bug? does any method to solve it?

armandabric:
This issue seems to have been fixed by #617. If not the case you could re-open the issue.

PR #617, opened 2021-09-27
feat: support more element types

Andarist:


rollup.config.js
```diff
@@ -36,12 +37,6 @@ export default {
     resolve({
       mainFields: ['module', 'main', 'jsnext', 'browser'],
     }),
-    commonjs({
-      sourceMap: true,
-      namedExports: {
-        react: ['isValidElement'],
-      },
-    }),
```
armandabric, line 44: Can you explain why you have to change that in the context of this PR?

armandabric (commented):
Wow! Thanks for this really god PR. I just have a question regarding the rollup configuration part.

rollup.config.js
Andarist: `commonjs` plugin is only needed when you actually want to **bundle** a commonjs dependency - you don't really want that with `react` since you are leaving it as external (like you should), so this was actually a "dead"/redundant config

armandabric:
This PR will fix #352 and is related to #617

Andarist:
@armandabric I've fixed the Flow issue 🎉  Also - I think that you have wanted to link another issue/PR since #617 is this one

src/parser/parseReactElement.js
```diff
@@ -12,12 +24,58 @@ import type { TreeNode } from './../tree';
...
+      return 'Lazy';
+    case isProfiler(element):
+      return 'Profiler';
+    case isStrictMode(element):
+      return 'StrictMode';
+    case isSuspense(element):
+      return 'Suspense';
+    default:
+      // this is "unknown" case and will most likely result in `[object Object]`
+      // it's fine though as we can't display anything meaningful anyway here
```
Andarist, line 75: Let me know if you prefer returning `"UnknownElementType"` here or something.

armandabric: I like your idea to return `UnknownElementType`. It will help to debug issue.

rollup.config.js
armandabric: Let's improve this!

Andarist: Hm, could you be more specific? maybe I've explained this poorly - so let me rephrase: the removed code that was not actually used and it wasn't doing anything here. And how this config now is defined is correct - from my point of view this change is "sound" and correct. There is no regression here so this can be merged "as is". Please let me know if this addresses your concerns - if not I can try to explain it further and in more detail.

armandabric: Ho! Sorry I just wanted to said _Let's change this!_ in the meaning of accepting your change with pleasure. Sorry for the wrong wording :smiling_face_with_tear:

armandabric (approved):
Thanks again for this PR! That was a long time running issue for the library :ok_hand:

xyy94813:
For `@storybook/addon-docs`, the scenario of converting a lazy component to a string is a little different from the needs of `react-dev-tools`.

```js
function MyComponent() {
  return 
}

const LazyComponent = React.lazy(async () => {
  return MyComponent
});

const root = React.render(
  <LazyComponent />
);

// for react-dev-tools
expect(jsxToString(root)).toBe(<Lazy />);  // first
expect(jsxToString(root)).toBe(<MyComponent />); // then

// for `@storybook/addon-docs` i will except
expect(jsxToString(root)).toBe(<Lazy(MyComponent) />); // then
```

Any suggestions about element to Source code?

Andarist:
Ah yes - I was wondering about this case but the problem here is that this `LazyComponent` doesn't have a name at all. We could try to infer it from the inner function **but** that is not an established pattern for `React.lazy`. We can't easily access the `MyComponent` name in the given example during the runtime.

PR #620, opened 2021-09-27
fix: improve perf by not stringifying React elements within `prettyPrint`

Andarist:
This has actually **fixed** an issue for me because `prettyPrint` was acting like an infinite loop. This is slightly odd because it implements `seen` tracking and shouldn't be prone to circular structures issues. What is even odder - after I've fixed this locally with `patch-package`, verified the fix, and reverted back so I could create a reduced test case for `react-element-to-jsx-string` then it stopped happening.

What I know is that it was looping through fiber data and its `return` property - which happened through a React element that was passed to `formatComplexDataStructure`. I'm not a fan of introducing code without knowing the root cause behind a problem and having a test case for the issue so I would totally understand if you wouldn't like to merge this in just yet. However, at the very least - this is a perf optimization (explained in the added comment) and for those things we usually don't write tests anyway so this could still be accepted "as is".

### DIFF
--- a/src/formatter/formatComplexDataStructure.js
+++ b/src/formatter/formatComplexDataStructure.js
@@ -18,6 +18,13 @@ export default (
   const normalizedValue = sortObject(value);
 
   const stringifiedValue = prettyPrint(normalizedValue, {
+    // `prettyPrint` first calls `filter`, then stringifies the value
+    // and only after that calls `transform` (with stringified value as 3rd argument - `originalResult`)
+    // since we don't care about `originalResult` for valid React elements and functions
+    // we can "skip" stringifying step by filtering out all of the keys using custom `filter`
+    // this acts as a perf optimization since React elements can hold fiber data and it can be a deeply nested structure
+    filter: currentValue =>
+      !isValidElement(currentValue) && typeof currentValue !== 'function',
     transform: (currentObj, prop, originalResult) => {
       const currentValue = currentObj[prop];
 


src/formatter/formatComplexDataStructure.js
```diff
@@ -18,6 +18,13 @@ export default (
   const normalizedValue = sortObject(value);
 
   const stringifiedValue = prettyPrint(normalizedValue, {
+    // `prettyPrint` first calls `filter`, then stringifies the value
+    // and only after that calls `transform` (with stringified value as 3rd argument - `originalResult`)
+    // since we don't care about `originalResult` for valid React elements and functions
+    // we can "skip" stringifying step by filtering out all of the keys using custom `filter`
+    // this acts as a perf optimization since React elements can hold fiber data and it can be a deeply nested structure
```
Andarist, line 25: Mentioned "steps" in the `prettyPrint`'s code can be found here:
https://github.com/Chris-Baker/pretty-print-object/blob/36a3b97482b81918593d472dc14b6f7b67c73338/src/index.ts#L175-L196

armandabric:
I'm not sur to understand when this append (in real life). But I understand the why this change could resolve an issue. 

You change propose to solve the issue at `prettyPrint` level. But as we are parsing the react element tree our-self could we detect the use case and act in a more explicit way?

Andarist:
> I'm not sur to understand when this append (in real life). But I understand the why this change could resolve an issue.

Ye, this is quite weird. I'm 100% sure that this has been happening to me because my browser tabs just became unresponsive and I had to kill them. I was also able to hit "Pause script execution" from time to time in the debugger and I was always ending up within this `prettyPrint`. And since I've been paused there I was able to inspect the call stack, the values passed around, and so on. Couldn't quickly pinpoint the issue as, like mentioned, it was all about the React's fiber structure and I'm not too familiar with how it is constructed. Then I've switched to patching it and the issue went away 😢 

> You change propose to solve the issue at prettyPrint level. But as we are parsing the react element tree our-self could we detect the use case and act in a more explicit way?

It's not as easy because `formatComplexDataStructure` receives an arbitrary object and a React element can be anywhere within it. So it makes sense to only try to handle a React element when the logic within `formatComplexDataStructure` encounters it. I see how this `filter` is a little bit weird - but that's mainly because this is the API/design of this `prettyPrint` package. 

This probably would look a little bit better if some other package would be used, like [`pretty-format`](https://github.com/facebook/jest/tree/fc97f3eb496e80b9cb69e05cd0246aaa3748eeb3/packages/pretty-format). But the package you are using is quite lightweight - so I definitely can see why it would be chosen over other alternatives.

armandabric:
Ok so it's a PR that to solve an issue that happen only append one after a browser crash. I think we will not fix it for that.

On the performance point of view, did you notice a notable performance improvement with this fix?

Andarist:
> Ok so it's a PR that to solve an issue that happen only append one after a browser crash. 

Note that the issue was persistent for a while - I was restarting my browser, using incognito sessions, etc and it all hand on that infinite loop.

> I think we will not fix it for that.

I respect that decision.

> On the performance point of view, did you notice a notable performance improvement with this fix?

Can't really say - I didn't measure it as I was focused on making it work in the first place. I might try removing the patch from my project and wait until the issue arise again - maybe then I would be able to debug it down and have a more compelling argument for merging this than "performance".

Issue #530, opened 2020-02-14
Support for overriding attribute values for displayName...


elliotgeno:
Not sure if this functionality is already available, but currently, if you provide another component to an attribute, the JSX that is rendered shows the value of the object instead of the displayName. For example:
`<Accordion.Toggle  as={Card.Header} >`
is rendered like:
`<Accordion.Toggle
      as={{
        $$typeof: Symbol(react.forward_ref),
        defaultProps: undefined,
        displayName: 'Card.Header',
        render: function noRefCheck() {}
      }}
    >`

I would like to only show the displayName in this scenario.
Is there a way to override the string the attribute returns?

If not, maybe we can make a slight adjustment to options.filterProps to work similarly to options.displayName?
If you provide a function, and that function returns a string that isn't a boolean, use that new string for the attribute value instead. 

Sound good?

elliotgeno:
In this case my filterProps object would look like this:
`filterProps: (value, key) => {
        if (key === "style") {
          return false
        }
        if (key === "as") {
          return value.displayName;
        }
      }`

dvnrsn:
This would be nice for really lengthy prop values as well. Say one has a long array of objects. It'd be nice to be able to render something like `array` or `[{}, {}]`

probablyup:
@Andarist would it be possible to add this functionality? kind of overloading `filterProps` to allow for prop value replacement or hiding it entirely

Andarist:
@probablyup note that I'm not a maintainer of this package - just fixed some issues here a few weeks back

as-zlynn-philipps:
Thanks for your work, @probablyup. I hope it gets merged. Although, it doesn't really address the need to format prop values on the fly as an escape hatch. I believe that would still be useful.

PR #682, opened 2021-11-29
feat: replace memoized/forwarded raw components as prop values

probablyup:
Fixes the root issue of #530


probablyup:
tag @Andarist

src/formatter/formatPropValue.js
```diff
@@ -53,6 +53,14 @@ const formatPropValue = (
     )}}`;
   }
 
+  // handle forwardRef and memo
+  if (isPlainObject(propValue) && propValue.$$typeof) {
```
probablyup, line 57: I tried to use `react-is` values here but it didn't seem to work for some reason.

Andarist: `react-is` is mainly designed to test **element** types but you want to check if a component is of a particular type.

You can still do it, although in a somewhat clunky way:

```suggestion
  if (isPlainObject(propValue) && (propValue.$$typeof === Memo || propValue.$$typeof === ForwardRef)) {
```

src/formatter/formatPropValue.js
```diff
@@ -53,6 +53,14 @@ const formatPropValue = (
...
+    return `{${propValue.displayName ||
+      propValue.type?.name ||
+      propValue.render?.name ||
+      'Component'}}`;
```
Andarist, line 61: I would recommend moving this logic to a shared file and reusing it here:
https://github.com/algolia/react-element-to-jsx-string/blob/af5b9d277fbd8de3a9a09d451ede6cfe5ee9768c/src/parser/parseReactElement.js#L34-L45

It would handle more cases, like for example `memo(forwardRef(Component))`

src/formatter/formatPropValue.spec.js
```diff
@@ -135,4 +135,32 @@ describe('formatPropValue', () => {
 
     expect(formatPropValue(new Map(), false, 0, {})).toBe('{[object Map]}');
   });
+
+  it('should format a memoized React component prop value', () => {
```
Andarist, line 139: I would recommend also adding some tests in the `index.spec.js` file

probablyup:
@armandabric would you be the right maintainer to tag?

probablyup:
@Andarist thank you for the review!

probablyup:
Why was this closed without comment?

armandabric:
It's an error, I closed a lot of dependencies upgrade PR

Fi1osof:
Is this will merged? I got error "react-element-to-jsx-string: Expected a React.Element, got `object`" while using ReactDOM.createPortal().

react-element-to-jsx-string@15.0.0

Issue #732, opened 2022-05-17
`useBooleanShorthandSyntax` strip false values


pascalduez:
The `useBooleanShorthandSyntax` option (which is on by default), strips `false` values.

While the behavior is fine for a `true` value: `<Foo bar={true}>` --> `<Foo bar>`, 
it's not for a `false` value: `<Foo bar={false}>` --> `<Foo>`, 
which is definitely not the same thing, the prop is removed completely.

armandabric:
Hi @pascalduez thanks for reporting this case. 

I see your point, the actual behavior came from the class component: we could set a default value on the for the props. We use there default value to know if the falsy boolean props should be shown. In that case it works as expected.

Here the class component we use to test this behavior:

https://github.com/algolia/react-element-to-jsx-string/blob/c9b726927de4aa147b901d2649951209a1a708b4/src/index.spec.js#L22-L29

And the actual test:

https://github.com/algolia/react-element-to-jsx-string/blob/c9b726927de4aa147b901d2649951209a1a708b4/src/index.spec.js#L855-L861

That being said, in the actual react word we could said that most of the components are functional one. We could try to change this behavior to show `false` props if no default value are set. I will give a try

pascalduez:
Hey @armandabric,

thanks for your reply.
I'm not sure to fully grasp, but this should be decorrelated from class vs functional components I guess.  
You can set "default props" on functional components as well, and in two ways.  
- Wit the `defaultProps` "static" property
- With default parameters

rmarquois:
I created a patch from @armandabric PR and it works.
Like @pascalduez said, I thinks library should handle the two ways to define defaultProps.

PR #733, opened 2022-05-18
feature: Show false props when not egal to default value

armandabric:
Fix #https://github.com/algolia/react-element-to-jsx-string/issues/732


idiglove:
Hey any update on when this can get delivered? It will fix Storybook issues, ref here https://github.com/storybookjs/storybook/issues/30704

PR #819, opened 2023-09-28
fix: Fix ESM distribution for node import usage

isaacs:
When loading this module with import() in Node, the CommonJS module was being loaded, resulting in {default:{default:fn}} instead of just {default:fn} as intended.

Because the types are only defined for CommonJS, this was further compounded, meaning that instead of doing this:

```ts
import reactElementToJSXString from 'react-element-to-jsx-string'
```

TypeScript users would have to do this:

```ts
const {
  default: { default: reactElementToJSXString },
} = (await import('react-element-to-jsx-string')) as unknown as {
  default: typeof import('react-element-to-jsx-string')
}
// wat.
```

- Add conditional exports for import and require, so that the ESM is served to import() users.
- Add a build script to place a package.json file in dist/esm, setting it to ESM mode for Node's loader and TypeScript's module resolution.
- Test that require() and import() receive matching objects, when loading from a mjs file in a folder outside this project.

With this change, `import()` in Node loads the ESM version, properly typed, without any hassle.

### DIFF
--- a/esm-dist.sh
+++ b/esm-dist.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Place a package.json so Node know that the dist/esm is ESM.
+echo '{ "type": "module" }' > dist/esm/package.json
+
+# Copy the types in there so TS knows it's ESM as well.
+cp index.d.ts dist/esm/index.d.ts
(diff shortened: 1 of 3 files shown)


armandabric:
@isaacs I will have look. It look likes the smoke test suite it not really well done for modern ESM build

PR #865, opened 2024-12-11
React 19

imjordanxd:
### Overview
Adds support for React 19. I don't have a lot of time to add this and would appreciate any help 🙏

##### New context rendering
In React 19, we can render a context as follows:

```tsx
const Context = React.createContext();
function Foo({ children }) {
	return <Context>{children}</Context>;
}
```

The React Element produced by this is **identical** (as far as I can see!) to the React Element produced if we were to render:

```tsx
const Context = React.createContext();
function Foo({ children }) {
	return <Context.Provider>{children}</Context.Provider>;
}
```

Because of this change, serialisation is broken. We don't know _how_ a context was rendered. Therefore, we can't produce a JSX string that accurately reflects the original JSX that was written. To support a wide range of React versions, we should serialise the new context rendering (`<Context>...</Context>`) as `<Context.Provider>...</Context.Provider>`. This isn't exactly a 1:1 serialisation but React will still treat the JSX the same.

Furthermore, the element type symbols have changed between React major versions. For example:

```tsx
// React 18:
<Context.Provider /> // Symbol.for('react.provider');

// React 19:
<Context /> /*or*/ <Context.Provider />  // Symbol.for("react.context");
```

This means we cannot bundle `react-is` as a dependency and allow a wide range or React peer dependencies. We should expect consumers of this package to provide `react-is` as a dependency which is aligned to their React version. This is a breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "prettier": "1.19.1",
     "react": "19.0.0",
     "react-dom": "19.0.0",
+    "react-is": "19.0.0",
     "react-shallow-renderer": "16.15.0",
     "rollup": "2.79.1",
     "rollup-plugin-babel": "4.4.0",
@@ -80,12 +81,12 @@
   },
   "peerDependencies": {
     "react": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0 || ^19.0.0",
-    "react-dom": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0 || ^19.0.0"
+    "react-dom": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0 || ^19.0.0",
+    "react-is": "^18.0.0 || ^19.0.0"
   },
   "dependencies": {
     "@base2/pretty-print-object": "1.0.2",
-    "is-plain-object": "5.0.0",
-    "react-is": "19.0.0"
+    "is-plain-object": "5.0.0"
   },
   "jest": {
     "setupFilesAfterEnv": [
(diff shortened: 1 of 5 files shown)


src/index.spec.js
```diff
@@ -7,7 +7,7 @@
 /* eslint-disable react/no-string-refs */
 
 import React, { Fragment, Component } from 'react';
-import { createRenderer } from 'react-test-renderer/shallow';
+import ShallowRenderer from 'react-shallow-renderer';
```
imjordanxd, line 10: this import crashes the test runner for some reason

armandabric: They cover some old pattern in class component. We could remove them and move on and shallow render is not a thing nowadays.

package.json
```diff
@@ -78,13 +80,13 @@
     "rollup-plugin-sourcemaps": "0.6.3"
   },
   "peerDependencies": {
-    "react": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0",
-    "react-dom": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0"
+    "react": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0 || ^19.0.0",
+    "react-dom": "^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1 || ^18.0.0 || ^19.0.0",
+    "react-is": "^18.0.0 || ^19.0.0"
```
armandabric, line 85: Let's stop this madness and use this major version to only support react 19

armandabric:
Thanks for your PR @imjordanxd I will cleanup a little a release a new major soon.

armandabric:
It's released https://github.com/algolia/react-element-to-jsx-string/blob/master/CHANGELOG.md#1700-2025-01-18

